PR #36, opened 2016-06-07
Add default Cocoa application menu for Mac apps. Fixes #35.

### LINKED ISSUE #35: No application menu for Mac apps
A common pattern for Mac applications is to have an application menu available to perform functions like quitting the application or viewing the 'About...' menu. I've attached a screenshot of the application menu from a sample application packaged with `py2app` to show what it normally looks like. 

screenshot 2016-06-06 16 59 13

In apps that are created with `pywebview` and packaged with `py2app` the application menu isn't accessible. The app name appears in its usual place but clicking it does not produce a menu. Is there a way to enable this functionality?

cuibonobo:
Allows Mac apps created with `pywebview` and packaged with `py2app` to follow the customary pattern of showing an application menu when the app name is clicked in the top menu bar. The new menu items are:

  * Services
  * _-----------------------_
  * Hide _[App Name]_
  * Hide Others
  * Show All
  * _-----------------------_
  * Quit _[App Name]_

The _[App Name]_ will only show in some menu items if `CFBundleName` is available in the bundle's `infoDictionary`. Otherwise it will be omitted.

The functionality for adding the menu and appending the app name were implemented as private methods to the `BrowserView` class so they don't change its public API.

I've also included a screenshot of the resulting menu on a dummy application bundled with `py2app`:

screenshot 2016-06-07 13 41 12

Fixes #35.

### DIFF
--- a/webview/cocoa.py
+++ b/webview/cocoa.py
@@ -102,6 +102,9 @@ class BrowserView:
             self.window.setCollectionBehavior_(NSWindowCollectionBehaviorFullScreenPrimary)
             self.window.toggleFullScreen_(None)
 
+        # Add the default Cocoa application menu
+        self._add_app_menu()
+
     def show(self):
         self.window.display()
         self.window.orderFrontRegardless()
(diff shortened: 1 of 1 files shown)


r0x0r:
Great work! Thank you.

PR #40, opened 2016-07-15
Add Gtk and Qt implementations of load_string.

ysobolev:
This adds load_string for the Gtk and Qt versions. Cocoa and Win32 versions throw NotImplementedErrors.

The added function has parameters:
`load_string(unicode_content, mime_type="text/html", encoding="utf-8", base_uri="")`

In the Gtk version, 'encoding' changes the encoding of the entire webview. This is not supported in Qt, but the parameter is passed in anyway (and ignored).

Example usage:

    # Python 2
    webview.load_string(u"日本")
    # Python 3
    webview.load_string("日本")`

If the string is not unicode in Python 2, it is converted for the user.

### DIFF
--- a/webview/__init__.py
+++ b/webview/__init__.py
@@ -103,16 +103,15 @@ def load_url(url):
     except NameError:
         raise Exception("Create a web view window first, before invoking this function")
 
-def load_string(content, mime_type="text/html", encoding="utf-8", base_uri=""):
+def load_html(content, base_uri=""):
     """
     Load a new content into a previously created WebView window. This function must be invoked after WebView windows is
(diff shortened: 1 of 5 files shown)


r0x0r:
Looks good. The only comment I have is regarding the function name. I think load_html is a more descriptive name than load_string.

ysobolev:
I will make the change. The reason it was load_string is because you can, in theory, pass in arbitrary resources (such as images). However, I was not able to get that to work.

ysobolev:
I replaced load_string with load_html. It simplified the encoding issues with Qt as well.

PR #54, opened 2016-10-09
Fix resizing width/height: 100% problem on GTK

klausweiss:
On GTK when body with/height is set to 100% in CSS after increasing the window size one could not resize content back to smaller size. I showed it on videos.

[before change](http://tinypic.com/r/iojhxs/9)
[after change](http://tinypic.com/r/ip3w2b/9)

What was the motivation for the changed line to be `add_with_viewport` instead of just `add`?

### DIFF
--- a/webview/gtk.py
+++ b/webview/gtk.py
@@ -55,7 +55,7 @@ class BrowserView:
         self.webview = webkit.WebView()
         self.webview.connect("notify::visible", self._handle_webview_ready)
         self.webview.props.settings.props.enable_default_context_menu = False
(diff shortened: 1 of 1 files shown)


r0x0r:
No motivation, just finding random code snippets and gluing them together. Software engineering at its best :D
Thanks for the fix.

klausweiss:
Great! Waiting for 1.2.2 ;)

r0x0r:
It is here :)
https://github.com/r0x0r/pywebview/releases/tag/1.2.2

Issue #45, opened 2016-08-29
"About" menu 


aaulich:
It would be great to have an "About" menu in the new default application menu.
This could display the app icon, the application name and some copyright and license infos.
E.g. a link to pywebview including its license could be added to this menu.

r0x0r:
A good suggestion, but with other features and suggestions in the pipeline the priority is low.
If you happen to implement it, a PR is welcome.

Issue #52, opened 2016-10-03
toggle_fullscreen() functionality


r0x0r:
Introduce a toggle_fullscreen() function that would toggle fullscreen mode on a currently active monitor. Currently the fullscreen mode on startup is activated on a primary monitor and there is no way to force it on another monitor.

r0x0r:
Fixed in master

PR #56, opened 2016-10-15
Added a standard Mac OS About menu

bastula:
May close r0x0r/pywebview#45

Advanced About dialog functionality can be found [here].(http://cocoadevcentral.com/articles/000071.php)


r0x0r:
Thanks for the PRs. #52 needs to be implemented for other platforms, as an actual toggle_fullscreen() function for Mac

Issue #62, opened 2016-11-09
Support for PyQt on Mac


josePhoenix:
There's a (perhaps narrow) use case where it would be beneficial to have pywebview on Mac try PyQt as well as PyObjC. The Anaconda scientific distribution does not (as far as I know) package PyObjC, but I am led to believe Qt is supported by the vendor as the anointed cross-platform GUI toolkit.

(I haven't yet tried fiddling with a local clone to see if it's as simple as adding a `try`/`except ImportError` for the Mac path in `__init__`...)

josePhoenix:
A brief unscientific test seems to indicate things "just work" when importing the Qt GUI on a Mac in the case where PyObjC is unavailable. Nice.

r0x0r:
Nice one. Actually I originally started a QT implementation on OSX  for some reason I don't remember right now.
It is unlikely I will work on this port, if any issues arise. But if you submit a PR, I will gladly accept it.

PR #63, opened 2016-11-10
Qt4/5 fallback for Macs with Qt and no PyObjC

josePhoenix:
See #62. This also uses ["install extras"](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) to specify which "flavor" of pywebview to install. If the dependencies are already present, installation proceeds just as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,11 +29,11 @@ To automatically fetch and install Python dependencies for your platform, instal
 
 On Windows (using WinForms with `pythonnet`):
 
-    pip install pywebview[windows]
+    pip install pywebview[winforms]
 
-On Mac (using the platform WebKit widgets via `pyobjc`):
+On Mac (using the Cocoa WebKit widget via `pyobjc`):
 
-    pip install pywebview[mac]
+    pip install pywebview[cocoa]
 
 On Linux (using GTK3 via `PyGObject`):
 
--- a/README.md
+++ b/README.md
@@ -20,8 +20,17 @@ Traktor Librarian: http://flowrl.com/librarian/
 
 # Installation
 
+If you have the [dependencies](#dependencies) installed, simply:
+
     pip install pywebview
 
+To automatically fetch and install Python dependencies for your platform, install with the appropriate ["install extras"](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies):
+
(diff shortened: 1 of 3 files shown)


webview/__init__.py
```diff
@@ -41,9 +41,17 @@ def _initialize_imports():
...
+            try:
+                import webview.cocoa as gui
+            except ImportError:
+                try:
+                    import webview.qt as gui
+                    logger.info("Using QT")
+                except ImportError as e:
+                    # Panic
+                    logger.exception("QT not found")
+                    raise Exception("You must have either QT or GTK with Python extensions installed in order to this library.")
```
r0x0r: It should say Cocoa and QT here

webview/__init__.py
```diff
@@ -41,9 +41,17 @@ def _initialize_imports():
 
     if not _initialized:
         if platform.system() == "Darwin":
-            import webview.cocoa as gui
+            try:
+                import webview.cocoa as gui
+            except ImportError:
+                try:
```
r0x0r: Maybe change it to the import_error flag to be inline with the rest of the function.

josePhoenix: 👍

webview/__init__.py
josePhoenix: You're absolutely right! Anecdote: I made this change twice because I screwed up line endings and git said every line was changed. It was correct in my first attempt... How do you feel about `.gitattributes` to enforce consistent line endings for history while letting contributors use platform line endings?

README.md
```diff
@@ -17,20 +17,36 @@ Apps created with pywebview. If you have built an app using pywebview, please do
...
+
+(Note that the same code is installed regardless; if more than one of the libraries below are available they will be tried in the order listed in this readme.)
+
+On Windows (using WinForms with `pythonnet`):
+
+    pip install pywebview[windows]
+
+On Mac (using the platform WebKit widgets via `pyobjc`):
+
+    pip install pywebview[mac]
```
r0x0r: Let's go with the framework name here as well (Cocoa) to avoid the confusion.

josePhoenix:
All right, all done addressing your comments. Cheers!

r0x0r:
Looks good. Merging.
Thank you for the contribution.

PR #67, opened 2016-12-26
Django App in examples

diegojromerolopez:
I'v added a Django App in examples. CherryPy Server Code is based on this [Moo Systems Tutorial](https://moosystems.com/articles/14-distribute-django-app-as-native-desktop-app-01.html).

To run this app, you only need to create a virtualenv using the requirements.txt file and launch the app using pywebview/examples/django_app/src/desktop_app_main.py.

The index URL has a input field and when launching from PyCharm the focus is in the parent terminal.


r0x0r:
This is too big and too specific for an example. Besides flask_example is designed to be a general purpose app skeleton (just replace server.py with Django code or such).

PR #78, opened 2017-02-07
Error when trying to use PYQT5

rittersport3:
Hi there,

I just created a Ubuntu 16.10, python 3.6 virtualenv environment and I couldn't use pywebview with the PyQT option.
On init import_error would always be False (because the error when importing the gui was caught in the module).

I am not sure this is the best way to solve this issue, but I am sharing what worked for me.

### DIFF
--- a/webview/__init__.py
+++ b/webview/__init__.py
@@ -67,7 +67,7 @@ def _initialize_imports():
                 logger.exception("GTK not found")
                 import_error = True
             else:
-                import_error = False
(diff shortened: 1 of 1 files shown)


r0x0r:
Thanks for pointing out to this issue. I have refactored the import logic in the https://github.com/r0x0r/pywebview/tree/import-refactor branch.
Would you please check out if this workaround resolves the issue?

PR #98, opened 2017-04-03
handle webview print requests in cocoa. fix #97

shivaprsdv:
The WebKit webview in macOS does not automatically handle the Javascript
window.print method. This fix let pywebview handle the print event via
`WebUIDelegate`'s [`webView:printFrameView:`](https://developer.apple.com/reference/webkit/webuidelegate/1387787-webview?language=objc) method.

PR #99, opened 2017-04-08
handle javascript alert method in cocoa

shivaprsdv:
A very basic solution. Follow up to #98

### DIFF
--- a/webview/cocoa.py
+++ b/webview/cocoa.py
@@ -58,14 +58,11 @@ class BrowserView:
             return nil
 
         # Display a JavaScript alert panel containing the specified message
-        def webView_runJavaScriptAlertPanelWithMessage_initiatedByFrame_\
-                (self, webview, message, frame):
-                    AppKit.NSRunningApplication.currentApplication().activateWithOptions_(AppKit.NSApplicationActivateIgnoringOtherApps)
-                    alert = AppKit.NSAlert.alloc().init()
-
-                    alert.setMessageText_(localization["cocoa.alert"])
-                    alert.setInformativeText_(message)
-                    alert.runModal()
+        def webView_runJavaScriptAlertPanelWithMessage_initiatedByFrame_(self, webview, message, frame):
+            AppKit.NSRunningApplication.currentApplication().activateWithOptions_(AppKit.NSApplicationActivateIgnoringOtherApps)
+            alert = AppKit.NSAlert.alloc().init()
+            alert.setInformativeText_(message)
+            alert.runModal()
 
         def webView_printFrameView_(self, webview, frameview):
             """
--- a/webview/cocoa.py
+++ b/webview/cocoa.py
@@ -57,6 +57,16 @@ class BrowserView:
         def webView_contextMenuItemsForElement_defaultMenuItems_(self, webview, element, defaultMenuItems):
             return nil
 
+        # Display a JavaScript alert panel containing the specified message
+        def webView_runJavaScriptAlertPanelWithMessage_initiatedByFrame_\
+                (self, webview, message, frame):
+                    AppKit.NSRunningApplication.currentApplication().activateWithOptions_(AppKit.NSApplicationActivateIgnoringOtherApps)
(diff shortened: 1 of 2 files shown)


webview/cocoa.py
```diff
@@ -57,6 +57,16 @@ class BrowserDelegate(AppKit.NSObject):
         def webView_contextMenuItemsForElement_defaultMenuItems_(self, webview, element, defaultMenuItems):
             return nil
 
+        # Display a JavaScript alert panel containing the specified message
+        def webView_runJavaScriptAlertPanelWithMessage_initiatedByFrame_\
+                (self, webview, message, frame):
```
r0x0r: No line break here

webview/cocoa.py
```diff
@@ -57,6 +57,16 @@ class BrowserDelegate(AppKit.NSObject):
...
+                    AppKit.NSRunningApplication.currentApplication().activateWithOptions_(AppKit.NSApplicationActivateIgnoringOtherApps)
+                    alert = AppKit.NSAlert.alloc().init()
+
+                    alert.setMessageText_(localization["cocoa.alert"])
```
r0x0r: The default window.alert in OSX has no title text, so this can be omitted or left empty

webview/localization.py
```diff
@@ -12,6 +12,7 @@
     "cocoa.menu.showAll": u"Show All",
     "cocoa.menu.quit": u"Quit",
     "cocoa.menu.fullscreen": u"Enter Fullscreen",
+    "cocoa.alert": u"Alert",
```
r0x0r: This can be removed

r0x0r:
My bad. It works just fine, I had a wrong branch selected. Just minor changes above and it is ready for merging.

webview/cocoa.py
shivaprsdv: [This](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) and [this](https://github.com/r0x0r/pywebview/blob/master/webview/cocoa.py#L135).
I usually break lines under 80 chars (I code in the good old Terminal `80x24` 🤓). And `cocoa` is full of these _long_ method names.

But if you insist, I shall rewrite it in a single line. Or is there some other _pythonic_ way?

r0x0r: Well the next line is very long anyway, so lets keep the style consistent. I tend to break a function definition line at parameters, breaking just after the name looks weird in my opinion.

webview/cocoa.py
shivaprsdv: You are right. In fact I wrote it this way after some consideration. It didn't feel right without a title because unlike the `window.alert` displayed by Safari, the native `NSAlert` we use _does_ usually have a title:

screen shot 2017-04-09 at 11 45 52 pm

_Without title (not pretty)_
screen shot 2017-04-09 at 11 49 53 pm

So I don't advocate making the title empty. Another option is to make the `message` itself the title (and omit the `informativeText` property), but that looks ugly and non-standard:

_This would be uglier if the message is something long_
screen shot 2017-04-09 at 11 50 41 pm

I just figured out it's better to _omit_ it as you said, because macOS shows `Alert` as the default title. So yes, if that fixes it, I'll make the changes.

webview/localization.py
shivaprsdv: Yes. This is unnecessary if we make the above proposed change. I shall commit once you follow up. 👍

webview/cocoa.py
shivaprsdv: My bad this time. I overlooked that second line. Oh my `cocoa`! 😄

I'll change it at once.

shivaprsdv:
[off-topic] Now when you merge, are you going do a _Squash and Merge_?

I'm still new to this stuff, just asking.

r0x0r:
I do not care that much about the git log purity, but squash and merge seemed like a fit choice here. Thanks for the contribution once again.

shivaprsdv:
Cheers. 👍

PR #103, opened 2017-04-19
Disable delete key back-navigation in Cocoa. Fixes #102.

### LINKED ISSUE #102: Pressing delete key causes back navigation
Pressing the Delete key on Mac causes the webview to navigate to the previous page.

Doesn't happen in Linux[GTK]. Don't know the case of Windows.

Seems undesirable in a desktop application.

shivaprsdv:
Prior to this commit, pressing the Delete key on the Mac keyboard caused
the WebView to navigate to the previous HTML page, which is undesirable
in a desktop application.

This fix makes use of the WebView's policyDelegate to refuse a back-
navigation actions if it was triggered by pressing the Delete key.
This is most convinient for two reasons:

1. We need not worry about Delete key press events that may occur
at other places (say inside text-fields).
2. Back navigation triggered explicitly via other means (like the
JavaScript history.back() method) can be allowed gracefully.

Note: attempts to capture the keypress via the WebView's keyDown method
fails because the parent NSWindow automatically assigns a subview called
WebHTMLView as the firstResponder to keyDown events, which handles
Delete key press events itself.

### DIFF
--- a/webview/cocoa.py
+++ b/webview/cocoa.py
@@ -85,6 +85,25 @@ class BrowserView:
 
             PyObjCTools.AppHelper.callAfter(printView, frameview)
 
+        # WebPolicyDelegate method, invoked when a navigation decision needs to be made
+        def webView_decidePolicyForNavigationAction_request_frame_decisionListener_(self, webview, action, request, frame, listener):
+            # The event that might have triggered the navigation
+            event = AppKit.NSApp.currentEvent()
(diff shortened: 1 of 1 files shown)


shivaprsdv:
The current last commit on this branch may be reset before merge.

r0x0r:
Great work. Thanks!

shivaprsdv:
Cheers! 🥂

PR #104, opened 2017-04-19
Disable backspace on Windows and OSX

r0x0r:


shivaprsdv:
@r0x0r Can't the `disable-backspace` branch be deleted now?

Issue #43, opened 2016-08-13
Support for multiple windows?


cuibonobo:
This feature will probably change the API significantly, but I wanted to at least document the need.

Essentially my idea would be that the `webview.create_window()` function returns a window handle or maybe even a `BrowserView` object that can be manipulated directly. That way you could instantiate multiple windows and destroy them individually.

Perhaps the app can quit when the last window is destroyed, or maybe there could be a 'master' window instance that can control all the other ones.

I'm kind of spit-balling but hopefully you get the idea. Could pywebview support multiple windows? Can you think of some other way to do it that doesn't involve a major API change?

r0x0r:
This is a good suggestion. One way to do it would be to leave the current API unchanged and introduce an additonal method create_window_multiple (or something like that) that would return a BrowserView object. 
This would require unifying the BrowserView objects across platforms and re-thinking a threading model. According to the current model, GUI is run in a main thread and backend logic in a separate daemon thread.

cuibonobo:
Ok, let me see if I'm thinking about this correctly: instead of `webview.create_window()` you would do something like `mywindow = webview.create_window_multiple()`, so now instead of, for example, `webview.load_url()` it would be something like `mywindow.load_url()`?

The problem is that in order for that to work, the `BrowserView` class shouldn't implement the singleton pattern. So instead of doing `BrowserView.instance = self` when we are initializing the `BrowserView`, we need to somehow keep track of the number of windows.

Maybe what can be done is instead of having `instance` as a class property, use a list called `windows`. That way `BrowserView.__init__()` can do something like this:

```python
self.windows.append(self)
```

This way we wouldn't actually need to add a `create_window_multiple` function: `create_window` can just return the `browser` object so it can be manipulated directly. If you don't want to work with objects, you can just call the module functions like normal. Following this plan, the module functions should access `windows[0]` instead of `instance`. For example:

```python
def load_url(url):
    BrowserView.windows[0].load_url(url)
```

r0x0r:
Yeah, something like this would do a trick. The main obstacle is that GUI frameworks dictate that GUI must be run in a primary thread. So changing a thread model and spawning multiple windows with create_window_multiple in separate threads would not work, unless multiprocessing is used.

What can be done instead is to make use of multiwindow capabilities of each framework. This would require going through documentation of each framework and figuring out how to do it separately for each platform. Ugh.

cuibonobo:
I was assuming that would be the case anyway :)

[...]

PR #109, opened 2017-04-30
[Cocoa] Improve handling of instances, give a title to child windows, and fix destroy

shivaprsdv:
* Fix broken examples
        - Thread called load_html too early. Adds delay, like in other examples.

* Close instance on calling destroy(), not quit app

* Give a title to the child windows
        - With the current design, there is no way to explicitly set the title of
the child windows (JS cannot set the title of the windows it opens). Web
browsers normally use the page title as the title for pop-up windows, so 
we follow suit.

### DIFF
--- a/webview/cocoa.py
+++ b/webview/cocoa.py
@@ -160,6 +160,15 @@ class BrowserView:
             new_browser.show()
             return new_browser.webkit
 
+        # WebFrameLoadDelegate method, invoked when the page title of a frame loads or changes
+        def webView_didReceiveTitle_forFrame_(self, webview, title, frame):
+            parent_window = webview.window()
+
+            # If the window has no title (e.g. it was opened by JS), use the
+            # HTML page <title> as the window title
+            if parent_window and not parent_window.title() and frame is webview.mainFrame():
+                parent_window.setTitle_(title)
+
 
     class WebKitHost(WebKit.WebView):
         def performKeyEquivalent_(self, theEvent):
@@ -256,7 +265,7 @@ class BrowserView:
             BrowserView.app.run()
 
     def destroy(self):
-        BrowserView.app.stop_(self)
+        PyObjCTools.AppHelper.callAfter(self.window.close)
 
     def toggle_fullscreen(self):
         def toggle():
--- a/webview/cocoa.py
+++ b/webview/cocoa.py
@@ -23,13 +23,15 @@ info["NSAppTransportSecurity"] = {"NSAllowsArbitraryLoads": Foundation.YES}
 
 class BrowserView:
     instances = []
-    active_instances = 0
     cascade_loc = Foundation.NSMakePoint(140.0, 0.0)
     app = AppKit.NSApplication.sharedApplication()
 
     class AppDelegate(AppKit.NSObject):
         def applicationDidFinishLaunching_(self, notification):
             BrowserView.instances[0].webview_ready.set()
+
+        def applicationShouldTerminateAfterLastWindowClosed_(self, sender):
+            return Foundation.YES
 
     class WindowDelegate(AppKit.NSObject):
         def display_confirmation_dialog(self):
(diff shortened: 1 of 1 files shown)


webview/cocoa.py
```diff
@@ -253,7 +262,7 @@ def show(self):
             BrowserView.app.run()
 
     def destroy(self):
-        BrowserView.app.stop_(self)
+        PyObjCTools.AppHelper.callAfter(self.window.close)
```
r0x0r: What would happen when you close the main window?

examples/multi_window.py
```diff
@@ -7,6 +7,8 @@
 """
 
 def load_html():
+    import time
+    time.sleep(1)
```
r0x0r: There is a semaphor which is set when webview is loaded, so timeout should not be needed. Not sure if the implementation is broken or not, though

webview/cocoa.py
shivaprsdv: You mean the window created first? It will just close. The app terminates only
when all windows of the app are closed.

This is in accordance with cuibonobo's [comment](https://github.com/r0x0r/pywebview/issues/43#issue-171017765), who suggested the idea to
begin with. The code that quits the app is in `windowWillClose` method of the
`WindowDelegate` ([here](https://github.com/r0x0r/pywebview/commit/d771625b42ac8917b6f3bdb61a1a686803809643#diff-2ec81f56c33197a32ca5ae4c092a6a2aL53)).

The current design makes no distinction between 'parent' or 'child' windows.
All windows, whether created by Python or JS, are 'children' of the app. You
want `pywebview` to give the initial window some special status? So that the
app terminates when it is closed? If this is the desired behaviour, it can be
implemented easily, I think. There is a `mainWindow` property to the app that
we can utilise. But the official documentation for the JS `window.open` makes
no such requirement. Try `window.open` in a browser, and the child window
stays open even after closing the browser's main window.

shivaprsdv: The current design makes no distinction between 'parent' or 'child' windows.
All windows, whether created by Python or JS, are 'children' of the app. You
want `pywebview` to give the initial window some special status? So that the
app terminates when it is closed? If this is the desired behaviour, it can be
implemented easily, I think. There is a `mainWindow` property of the app that
we can utilise. But the official documentation for the JS `window.open` makes
no such requirement. Try `window.open` in a browser, and the child window
stays open even after closing the browser's main window.

examples/multi_window.py
shivaprsdv: Then the bug is still on I think. I added the delay because after adding multi-
window support, I occasionally get this:

screen shot 2017-05-03 at 10 09 17 pm

It's just a tiny delay to be on the safe side, can't hurt, can it?

shivaprsdv: The actual fault is in my implementation of `get_key_instance`. It returns
`None` when no window has focus. I did it on purpose, thinking that the
way things are, no one can get back into the Python interpreter without
focusing and interacting with any pywebview window. This one is an
exceptional case, so instead of returning `None`, may be we should
raise an `Exception`?

This is not a problem when the app is bundled into executable, I have
verified it on Mac. It's only when running from the command-line that
this pops up. I think adding this delay is the best way to get away with
it. After all it's just a demo. 😉

r0x0r: Ok, let's get rid of example timeouts for now and this PR can be merged. 

Also let's go with the original approach without get_key_instance(). Window management can be done via Javascript for now.

shivaprsdv:
Ok, it was _so_ stupid of me to implement `get_key_instance` this way. 🤓 

I have got rid of example timeouts as suggested. But before throwing my
`get_key_instance` out of the window, I have decided to give it one more
chance because the original approach is equally bug-prone. What if some
one tries to call the API after the initial window is closed? All crazy things
happen. I have checked.

I have thoroughly tested it this time. Please checkout the edits and tell me
if you are still unsatisfied. Cheers. 👍

r0x0r:
If I understand correctly get_key_instance returns either a window currently in focus or the last opened window. Is this correct? 
What is exactly the use case of programmatically destroying / changing the url of a currently focused window? What if you want to close a window from another window? With this approach it would not be possible.

shivaprsdv:
Yes, precisely.

Correct again, good point. This happened because I was only blindly trying to
adapt the API to the current scenario with minimum modifications to the API
itself. I couldn't also find a good method to incorporate your suggestion of
using an instance_id, so `get_key_instance` was my best bet. 

But finally, I have found a way to give each window an instance_id that we can
use to cross-refer the window from JS or Python. Hopefully all API methods can
be made to point to any window using this identifier. Phew!

The concept is fairly simple, but right now it's a WIP and should be a PR on
its own. I'm cleaning up *this* PR to include only the changes you have approved
so far. So `get_key_instance` goes back to the shape it has in 6cbd959. For the
rest the currently reviewed diff remains almost unchanged.

This also means the examples will remain broken. This should be only temporary;
they ought to be automatically fixed by the proposed instance_id PR.

Issue #90, opened 2017-03-06
Loading indicator


rxng:
Hi 

Is there a way to add a loading indicator or possibly even a progress bar to this?

R

r0x0r:
To clarify the question, do you mean a loading indicator to be displayed before WebView is loaded?

rxng:
Yes, that is right.

r0x0r:
Definitely doable, but it is a fairly big effort, as the loading indicator
must be implemented separately for each platform. Creating an animation in
Win32 or Cocoa doesn't sound very appealing.

r0x0r:
I have given it a more thorough thought and imo the way to do it would be introducing two new parameters to the create_window function: background_color and placeholder_image. The first is the background color of the container and the latter would be a path to an image file displayed during loading (with gif support). 
Help is welcomed.

shivaprsdv:
I believe what you just described👆 is an HTML5/CSS3 thing. This is something
the web dev should implement themselves, should they need such eye-candy.
There are any number of websites that does this: load a lightweight page with a
progress indicator, load the main page in the background using AJAX, and switch
once load is finished. Use HTML5 `<progress>` element, jQuery, etc.

What @rxng asked is, I believe, a progress bar like the one shown by browsers,
that would animate on each load. But that would spoil the native-application UX.
So IMO this should not be a part of the API. If web devs intend on loading heavy
pages/internet resources, and want to avoid users staring at blank pages/stuck
pages, they should handle it themselves.

_My two cents._ 🙂

r0x0r:
The loading indicator in question is intended to be displayed before webview is loaded. Currently it a  white background is displayed for a second or two before any HTML content is displayed. Giving an ability to display an image as a loading indicator would make possible to display the same image before the webview is loaded and again the same image in the application itself.

shivaprsdv:
What about, display the window only after the content is fully loaded?

I think this can be implemented without much effort in the source.

r0x0r:
This could do the job, but in this case a loading splash screen would be needed. Progress indication is always better than just leaving the user hanging.

shivaprsdv:
Correct. At least on Mac this shouldn't be a problem, there is the default
'dock-icon-jumping' thing.

r0x0r:
I have been thinking the typical case would be the background colour of the application and a loading animation (a spinner or such). Anyhow both background colour and image would be left up to the user (no defaults). 

[...]

PR #110, opened 2017-05-15
[Cocoa] Add background color 

shivaprsdv:
Also added an example script.

We use another webview as the placeholder till the webview is loaded.
Param background_color and loading_image become the CSS background-color
and background-image of the webview's <body> respectively. This is the
easiest hack that is also portable across platfroms. :)

### DIFF
--- a/webview/cocoa.py
+++ b/webview/cocoa.py
@@ -5,6 +5,7 @@ Licensed under BSD license
 http://github.com/r0x0r/pywebview/
 """
 import threading
+import re
 
 import Foundation
 import AppKit
@@ -124,6 +125,13 @@ class BrowserView:
             # Normal navigation, allow
             listener.use()
 
+        # Show the webview when it finishes loading
+        def webView_didFinishLoadForFrame_(self, webview, frame):
+            # Add the webview to the window if it's not yet the contentView
+            if not webview.window():
+                BrowserView.instance.window.setContentView_(webview)
+                BrowserView.instance.window.makeFirstResponder_(webview)
+
 
     class WebKitHost(WebKit.WebView):
         def performKeyEquivalent_(self, theEvent):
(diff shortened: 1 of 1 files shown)


r0x0r:
Doesn't using another webview defeat the purpose? The idea is to display a
loading indicator as quickly as possible until the actual webview is
loaded. Having another webview will just introduce another delay.

shivaprsdv:
Uh. Isn't the delay associated with loading an HTTP request? This doesn't
load any request, just load a local image file. I don't see any white screen.

Wait let me check.

shivaprsdv:
Stupid me. You were right. There's a brief white flashing. I didn't notice
before because I was running all the scripts from a fullscreen terminal!

And here I was thinking I had found an ingenious hack. 😕 Let me check
if there is a workaround.

shivaprsdv:
Okay, let us get the better out of it. The delay for displaying the placeholder
webview must be very small even on slower machines (unless the loading_image
is something extreme-super huge, which is unlikely). So the window can be safely
held back till it finishes loading. We switch to the main webview once it finishes
loading (which may typically take longer).

It's not as easy with the native Cocoa API's to display animated GIFs and stuff.
Must be the same with other platforms as well, right? This would be a nice hack
if we could pull it off..

shivaprsdv:
Now this revision doesn't flash the white background.

r0x0r:
I have tested with [Next](http://flowrl.com/next) with both OSX and Windows. With OSX the loading time is barely noticeable (a second or so), while on Windows the white background stays for several seconds.
I would approach this problem by implementing a background color for a webview window in the first place. That should be not be too hard. This way the loading progress could be implemented in a gradual manner: Background color of the application (set by pywebview) -> Loading animation of the application (set inside the application) -> the fully loaded app.

shivaprsdv:
Modified PR according to [issue comment](https://github.com/r0x0r/pywebview/issues/90#issuecomment-301610310).

r0x0r:
The background color is still visible for the title bar after the webview is loaded. Strangely enough the problem does not occur when a breakpoint is set in the webView_didFinishLoadForFrame_ handler. Also the background color disappears when the window is dragged to another monitor. 
screen shot 2017-05-17 at 14 47 07

shivaprsdv:
Fixed.

Also [fixed](https://github.com/r0x0r/pywebview/pull/110/files#diff-2ec81f56c33197a32ca5ae4c092a6a2aR411) hex being converted to wrong color space (`SRGB` instead of `Calibrated`).

Issue #159, opened 2017-12-11
Option to set the title tags as the name of the window


guiscaranse:
It would be amazing if we could have the option to set the title of the window by using only the `<title>` tag on the HTML loaded.

r0x0r:
I am not sure how I feel about this. There are several options to tackle this issue
1) Make `title` parameter of `create_window` optional and update the title based on the title of the loaded page. I am not too fond of this approach, as changing the app title on the fly is not good UX in my book. Browsers for example prepend the title with the browser name, but in case of pywebview this cannot be done.
2) Introduce `set_title()` function and leave the logic of updating the title to developer
 
Input is welcomed

guiscaranse:
Both of the solutions are good, even tho I prefer the first approach. In my case changing the title of the window is important because developers can implement things like "notification counters" as many apps currently use.
`set_title()` is good too because developers can make callbacks from the javascript to python, so updating the title whatever they want can be useful.

r0x0r:
Let's go with the `set_title()` function and leave automatic update to end developers.

r0x0r:
I created an implementation skeleton + Cocoa implementation in the `set_title` branch

r0x0r:
The rest is implemented as well. I am yet to test it on Windows, but otherwise it is done.

r0x0r:
PR is here https://github.com/r0x0r/pywebview/pull/167
I will keep it open for a couple of days in hope of comments and suggestions,

PR #167, opened 2018-01-26
 #159 Implementation of set_title() function

r0x0r:
+ Fix a js_bridge exception on GTK
+ Bonus refactoring

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,9 @@ Create a new WebView window. Calling this function will block application execut
 
 These functions below must be invoked after webview windows is created with `create_window()`. Otherwise an exception is thrown.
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -112,6 +112,9 @@ Create a new WebView window. Calling this function will block application execut
 
 These functions below must be invoked after webview windows is created with `create_window()`. Otherwise an exception is thrown.
 
+- `webview.set_title(url)`
```
guiscaranse: Hey @r0x0r is it really `url` as parameter? Shouldn't it be a text or string?

r0x0r: Changed

PR #172, opened 2018-02-02
fixed some problems with PyQt versions

adbenitez:
please check lines 146 and 294, probably you really want to use > instead of >=
and line 252 and 301 probably most to be >= instead of >
I am not familiar with Qt so I can't guess, any way I will investigate about it if I have time.


shivaprsdv:
@adbenitez Lines 146 and 294 are okay, the QWebChannel interface is available
since QT 5.4. You are right about 252 and 301; till now we have not spotted any
issue with them, but we'll keep an eye on it. Thanks.

Issue #128, opened 2017-08-31
Fix tests


r0x0r:
Both Travis and AppVeyor tests fail at the moment. Travis seems to have a configuration issue. AppVeyor fails are semi-random and seem to do with race conditions. I could not reproduce AppVeyor failures locally.

shivaprsdv:
Tests are now passing on both. Seems to be the result of fbe56e00b41d34cebcb6c489b65a5771d564fa6c.
Or is the issue solved?

r0x0r:
Disabling 32-bit builds certainly helped, but there are still random fails from time to time. Still no idea what causes fails, as I could not reproduce this issue locally. Might be something fundamentally wrong how tests are run, for all I know.

shivaprsdv:
Another thing, all tests are failing locally on macOS Sierra. On googling, it looks
likely an OS issue. Still thought I'd report.

shivaprsdv:
Currently OSX builds on Travis are failing with a `sudo: pip: command not found`
error. This seems to be different from previous issues. Can we check if it is a bug
from their side?

r0x0r:
Looks like their OSX configuration changed and pip is no longer on the path. I will investigate.

shivaprsdv:
- [ ] [macOS] Fix mouse helper function

r0x0r:
It should be noted that the mouse helper is sort of a hack. The reason it exists is because Coca events do not fire unless mouse is moved. I don't know why Cocoa has this behaviour and whether there is a better solution than moving a mouse cursor programmatically, but at least it had got the job done until now (I blame a recent OSX update for this).

r0x0r:
The current GTK implementation is broken on master and tests do not catch that. It throws an exception in the GUI thread, which is reported to the console, but is not caught by the catch block in the test runner. I fixed the bug in `set_title` branch, test runner is yet to be addressed.

shivaprsdv:
> I blame a recent OSX update for this
Tests are running fine for me locally (macOS High Sierra 10.13.2, built in Python 2.7)
Which update is causing the issue? Or is it a Travis only issue?

r0x0r:
I think it might be a (High) Sierra issue. Anyhow I downgraded the OSX image on Travis to 10.11 and suddenly the tests pass. Let's keep it this way for the time being.

shivaprsdv:
I think the `assert_js` function is badly formulated. It's leading to infinite loops
on Linux.

[...]

PR #183, opened 2018-02-22
Partially fix Travis CI tests

shivaprsdv:
With this fix all remaining Travis test fails would boil down to the `assert_js`
function (#128).

### DIFF
--- a/webview/js/api.js
+++ b/webview/js/api.js
@@ -20,12 +20,12 @@ window.pywebview = {
     },
     _bridge: {
         call: function (func_name, params) {
-            if (window.external) {
-                return window.external.call(func_name, params)
-            } else if (window.qt) {
+            if (window.qt) {
                 new QWebChannel(qt.webChannelTransport, function(channel) {
                   channel.objects.external.call(func_name, params)
                 });
+            } else if (window.external) {
+                return window.external.call(func_name, params)
             }
         }
     },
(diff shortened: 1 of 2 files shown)


webview/js/api.js
```diff
@@ -20,12 +20,12 @@ window.pywebview = {
     },
     _bridge: {
         call: function (func_name, params) {
-            if (window.external) {
-                return window.external.call(func_name, params)
-            } else if (window.qt) {
+            if (window.qt) {
```
r0x0r: Why does the order of ifs matter here?

webview/qt.py
```diff
@@ -24,7 +24,7 @@
     from PyQt5 import QtCore
 
     # Check to see if we're running Qt > 5.5
-    from QtCore import QT_VERSION_STR
+    from PyQt5.QtCore import QT_VERSION_STR
```
r0x0r: Is this semantic sugar or is there more to this change?

webview/js/api.js
shivaprsdv: Because in some recent versions of QT, `window.external` is present, without
pywebview creating one. If we check for it first, we will wrongly assume that it
is part of our JS API object. Please read 5883780 commit message.

webview/qt.py
shivaprsdv: There is. Again, I don't exactly know why this is needed, but sometimes Python 3
doesn't recognise `QtCore` as a module and throws a false `ImportError` when
QT5 is present. This seems to pass both Python 2.x and 3.6.

r0x0r:
Ok to merge

PR #203, opened 2018-03-23
Fixed js not importing issue

jjgalvez:
this should fix the JS folder not loading. Its tested on linux and mac. sorry I don't a PC that  I could test it on
Jose

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import platform
 from setuptools import setup
 
 package_data = {
-    'webview.js': ['*.py']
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools import setup
 
 package_data = {
```
shivaprsdv: If this indeed works and get approved, you could simply remove this assignment
altogether instead of commenting it out. @jjgalvez

PR #204, opened 2018-03-23
Examples clean up

adbenitez:
Some clean up and just added a bit of comments, and use of `webview.webview_ready()` instead of `time.sleep` just to be safe.


shivaprsdv:
Hi, it seems to be your best intentions, but I don't really think we need to
call `webview_ready()` before any of the API functions, not at least "to be
safe". Because what it does (call `_webview_ready.wait(timeout)`), each
function does on its own, before passing on the call to the platform-specific
scripts. I think Roman wrote it to help with the tests, I have not seen it used
any where else.

So the `time.sleep` in the examples, I guess, is to give the person running
the examples some time to see what was happening. e.g. if the url changed
immediately after creating a window, it would be hard seeing it. Of course,
I can only guess, for Roman wrote them to begin with.

The cleaning-up-stuff seems good. But I suggest you clean up your commit
tree as well; it has got changes you made before that has been merged. 🙂

adbenitez:
I am new with git and don't have previous experiences collaborating in any project other than this one :(
I most to investigate how to clean up my commit tree.
What you see about `webview_ready()` is true partially because every webview method like `evaluate_js` call `_webview_ready.wait(5)` which will will give an error if the window is not ready in 5 seconds, I now this is practically impossible but I prefer to be 100% safe. Anyway I understand this is unnecessary for the examples, I will revert it :)
Sorry, I just wanted to help with the little free time I have right now.

shivaprsdv:
Please don't worry, I remember being even messier when I started out (had
to completely wipe my local pywebview repo at least twice before cloning
from scratch again 😄); and I am no expert in git either, that I couldn't have
figured out a way to clean up your commit tree, if I had been in your place.
It's okay, Roman doesn't care much of his commit log; and anyway, there is
the Squash and Merge option on our side (it squashes all of your commits
into a single commit before merge, so you don't need to bother clean up):
screen shot 2018-03-24 at 1 55 31 am

And your help is _more than welcomed_. That's what this project needs. It's
been mainly Roman alone before I started dropping in bits and pieces, until
he gave me maintainer's role. He'd be grateful for the "little free time" you
give for pywebview. And you can do better: Issues are urgent ones, you can
see if you can solve some of them (I see you've been opening many recently,
and that is good contribution too!). Or if you are good at commenting code,
you could help with the README, or wiki.

Sorry if I sounded rude. Thanks for your time. Cheers. 🙂

PR #205, opened 2018-03-23
Examples clean up

adbenitez:
Just some code clean up


lwerdna:
How does this example work when it still uses js_api and debug keyword keyword arguments that don't exist in my /usr/local/lib/python2.7/site-packages/webview/__init__.py? Pip says I'm on version 1.8.

PR #207, opened 2018-03-26
Refactor API calls to remove duplicate code

shivaprsdv:
@r0x0r This is a change at the core, hence the PR.

### DIFF
--- a/webview/qt.py
+++ b/webview/qt.py
@@ -232,13 +232,12 @@ class BrowserView(QMainWindow):
             js_result['result'] = None if result is None or result == 'null' else result if result == '' else json.loads(result)
             js_result['semaphore'].release()
 
-        escaped_script = 'JSON.stringify(eval("{0}"))'.format(script)
 
         try:    # PyQt4
-            result = self.view.page().mainFrame().evaluateJavaScript(escaped_script)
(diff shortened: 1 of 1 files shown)


r0x0r:
This is too implicit in my opinion. How about doing the same thing using decorators? There could be a decorator for exception handling.

shivaprsdv:
Er, what is too implicit? And I don't know much about decorators. 🤔

r0x0r:
Decorators are dead easy, even though the concept might sound scary. In fact `_call_api` can be recycled for a decorator.  
Here is a good intro on decorators https://realpython.com/primer-on-python-decorators/

shivaprsdv:
Okay, I've got two things:

1. I read that article about decorators and got it. Rewrote the refactor to
use them.
2. The last time you pulled from GitHub, seems things didn't go very well.
Stuff like the Readme about `webview.start` came back. So I set out to
straighten out the commit history a bit. Succeeded, but messed up this
PR in the process. I'll open a new one.

I have done a rebase, you'll have fetch and hard checkout (when things
like this happen, I tend to delete the local branch and checkout afresh).
Pulling won't help; I am sorry for this inconvenience, but otherwise you
would have had to repeat a lot of changes you made upstream.

PR #208, opened 2018-03-29
[All] Refactor API calls to remove duplicate code

shivaprsdv:
Rewrote the original #207 to use decorators.

### DIFF
--- a/webview/__init__.py
+++ b/webview/__init__.py
@@ -20,6 +20,7 @@ import re
 import sys
 from threading import Event, Thread, current_thread
 from uuid import uuid4
+from functools import wraps
 
 from webview.util import base_uri, parse_file_type, escape_string, transform_url, make_unicode, escape_line_breaks, inject_base_uri
 from .js import css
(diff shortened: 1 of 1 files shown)


r0x0r:
Good work! One thing, maybe give `api_call` a more descriptive name, like `handle_exceptions` or something along these lines?
Other that, ok to merge.

shivaprsdv:
Hm, isn't `api_call` is more descriptive than `handle_exceptions`? The latter
feels a bit generic and vague; it can mean handle all sorts of exceptions, while
the purpose of the decorator is to handle a very small subset of them, _specific_
to API calls. It reads like: "what follows is an `api_call`"; if someone ask what
an `api_call` is or does, well it is spelled out in its docstring.

But if you insist, it's okay with me, I shall change it.

shivaprsdv:
Another thing, I have modified the commit to prefix `api_call` with an underscore
to mark it as private.

r0x0r:
Let's leave it as _api_call, until a better name manifests itself.

Issue #213, opened 2018-04-13
add support for OpenBSD 6.x


hucste:
### Specification
- Platform: OpenBSD
- Version: 6.2, 6.3

### Description
Cant install! 

See this:
```
$ python3 -m pip install -U pywebview 
Collecting pywebview
  Using cached pywebview-1.8.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-dn5vt0ol/pywebview/setup.py", line 29, in <module>
        extras_require=extras_require,
    NameError: name 'extras_require' is not defined
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-dn5vt0ol/pywebview/

```

It seems necessary to modify your ''setup.py'' to add an extras as practically Linux. 

We have modules python3:
- py3-gobject3
- py3-qt5

;)

vv221:
I think the main issue is with the structure of the 'extras_require' definition: an 'if/elif/elif' block without an 'else' statement is prone to error and will break on some exotic configurations.

A simple 'else' statement interrupting the script execution and displaying an error like "Your system does not seem supported. Supported systems are: …" would avoid confusion.

shivaprsdv:
Good point. May be @r0x0r wrote it this way because pywebview explicitly
supports only those there platforms. But I can't see why a BSD with Gtk/Qt
be denied the opportunity to experiment.

@hucste Meanwhile have you tried if it works by cloning the repo locally?

hucste:
@shivaprsdv no, i'm going!

hucste:
Ok, i installed py3-qt5! It seems ok... 
py3-gobject3 was intalled, previously... but, the script cant detect it!

----

I add "or platform.system() == 'OpenBSD':" into '\_\_init\_\_.py' at line 78. 
Egual, into 'qt.py'... into lines 34 and 158, because i dont know the reason but even QT version is 'QT Version: [5, 9, 1]', it cant import this: 
`from PyQt5.QtWebEngineWidgets import QWebEngineView as QWebView`

> ModuleNotFoundError: No module named 'PyQt5.QtWebEngineWidgets'

----

To resume, if wrote '\_\_init\_\_.py' script, line 78 as:
`       elif platform.system() == 'Linux' or platform.system() == 'OpenBSD':`

[...]

PR #215, opened 2018-04-14
Add support PyQt5 for OpenBSD 6.x!

hucste:
Add new support for OpenBSD 6.x with PyQt5, as reply for #213!


webview/qt.py
```diff
@@ -30,7 +31,7 @@
     from PyQt5.QtCore import QT_VERSION_STR
     _qt_version = [int(n) for n in QT_VERSION_STR.split('.')]
 
-    if _qt_version >= [5, 5]:
+    if _qt_version >= [5, 5] and platform.version() == 'Linux':
```
r0x0r: This should probably be `and platform.version() != 'OpenBSD'`, as QT can be used on OSX as well.

webview/qt.py
```diff
@@ -154,7 +155,7 @@ def __init__(self, uid, title, url, width, height, resizable, fullscreen,
         self.evaluate_js_trigger.connect(self.on_evaluate_js)
         self.set_title_trigger.connect(self.on_set_title)
 
-        if _qt_version >= [5, 5]:
+        if _qt_version >= [5, 5] and platform.version() == 'Linux':
```
r0x0r: Same here

webview/qt.py
hucste: Ok, i attempt! ;)

hucste: Not run! :(

hucste: But this run:
```if _qt_version >= [5, 5] and (platform.version() == 'Linux' or platform.version() == 'Darwin'):```

One idea?!

Comment (hucste) [reply webview/qt.py]: OK, it seems i've a problem with the test: 
```
$ python3 
Python 3.6.2 (default, Oct  2 2017, 09:52:18) 
[GCC 4.2.1 Compatible OpenBSD Clang 4.0.0 (tags/RELEASE_400/final)] on openbsd6
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.version()
'GENERIC.MP#7'
>>> print(f'OS: '+ platform.version())
OS: GENERIC.MP#7
```

```
$ python 
Python 2.7.14 (default, Oct  2 2017, 09:52:26) 
[GCC 4.2.1 Compatible OpenBSD Clang 4.0.0 (tags/RELEASE_400/final)] on openbsd6
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> print 'OS : %s' % platform.version()
OS : GENERIC.MP#7
```

I think that's not normal!

hucste: Ok, it's `platform.system()` to detect and not `platform.version()`!

hucste: OK, this corrected with https://github.com/r0x0r/pywebview/pull/215/commits/a780407d84bec60a98acf51b16529cb2b6d6d9e3! 

So good for you?!

r0x0r: Yes, it is fine now. Thanks!

PR #221, opened 2018-05-08
webview/lib/*dll files should copy over.

mdigi:
I think this should work now for copying the lib folder for Windows.


mdigi:
Copying the *dll files work without this commit.

PR #224, opened 2018-05-12
Qt.py fix

adbenitez:
`_initialize_imports` was expecting an `ImportError` while `webview.qt` was rising an `Exception`


adbenitez:
also using tuple instead of list for qt versions

webview/qt.py
```diff
@@ -51,21 +51,12 @@
...
-    except ImportError as e:
-        _import_error = True
-    else:
-        _import_error = False
+    from PyQt4 import QtCore
+    from PyQt4.QtWebKit import QWebView, QWebPage, QWebFrame
+    from PyQt4.QtGui import QWidget, QMainWindow, QVBoxLayout, QApplication, QDialog, QFileDialog, QMessageBox, QColor
 
-if _import_error:
-    raise Exception('This module requires PyQt4 or PyQt5 to work under Linux or *BSD.')
```
r0x0r: Should this raise ImportError instead? If I read the code correctly, this change will just report that PyQt4 is not installed misleading that pywebview works only with PyQt4

adbenitez: yeah, I am aware of that but I also found throwing an `ImportError` in a line where there are no `import` statement a bit messy, I was trapped between this two options :( but to make `qt.py` work as the counterparts and since this error only matter if  pywebview don't find any gui toolkit and _initialize_imports will report "you must have XXXXX, and PyQt4 or PyQt5" I think is not a big issue, anyway we can make a logger.warning when PyQt5 fail to load to be even more explicit.

adbenitez: or logger.error

adbenitez: or logger.exception

r0x0r: I see no problem with raising an ImportError. Rethrowing an exception is a common practice in Python and technically the rethrow code could be inside the first catch. `import_error` is there to avoid code nesting.

adbenitez: We can also replace the line 48 of my modification to `qt.py` (`logger.debug('PyQt5 or one of dependencies is not found', exc_info=True)`) with `logger.exception('PyQt5 or one of dependencies is not found')` 
but I analyzed better and you are right there are no need to introduce this changes if replacing `Exception` with `ImportError` will do what we need, but please consider replacing the message `'This module requires PyQt4 or PyQt5 to work under Linux or *BSD.'` with something like `"No module named 'PyQt4' or 'PyQt5'"` or `"Could not load PyQt4 or PyQt5` looking forward to Qt support on Windows. I will be closing this pull request :)

PR #227, opened 2018-05-14
Initialize import enhancement

adbenitez:
Aiming to make more easy to add new back-ends, also making possible to use gtk in qt-based OS, while keeping compatibility with the old `use_qt` and `use_win32` properties, also trying to fallback into other supported back-ends if the preferred one(s) aren't found (ex. trying gtk if QT isn't installed when `PYWEBVIEW_GUI=QT`  when platform.system() == 'Linux')


adbenitez:
I really most to do something with my commit history, I have followed https://help.github.com/articles/syncing-a-fork/ I don't know what I have made wrong :(

PR #234, opened 2018-06-04
fix #232 : toggle_fullscreen doesn't work as expected

### LINKED ISSUE #232: toggle_fullscreen doesn't work as expected
### Specification
- Platform: winforms
- Version: 2.0.3

### Description

I just ran the example of todos. It came error and existed unexpectedly,when I clicked the to *full screen* button twice. The window didn't resize as beginning and just stop work. Should I make a PR for this?

lt94:
a few days I had created an issues [toggle_fullscreen doesn't work as expected](https://github.com/r0x0r/pywebview/issues/232), and i had a try to fix this problem by myself. It seemed that no matter variable *self.is_fullscreen* is, variable *screen* is always expected to obtain

### DIFF
--- a/webview/winforms.py
+++ b/webview/winforms.py
@@ -197,13 +197,11 @@ class BrowserView:
 
         def toggle_fullscreen(self):
             screen = WinForms.Screen.FromControl(self)
-
             if not self.is_fullscreen:
                 self.old_size = self.Size
                 self.old_state = self.WindowState
                 self.old_style = self.FormBorderStyle
                 self.old_location = self.Location
-
 
                 self.TopMost = True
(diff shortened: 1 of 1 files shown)


r0x0r:
I have no idea why it works, but it does. Good work.

lt94:
So do I, but it works.  That's why i love coding so much,there are always something i don't know

lt94:
Any ideas about test failed?

shivaprsdv:
AppVeyor tests seem to be failing even on master, so your PR has not likely
caused the issue. @lt94

lt94:
Thanks a lot@shivaprsdv

PR #242, opened 2018-08-09
Update __init__.py

wormius:
When webview.load_css(css, child_uid)  it's used, the interpreter throws:

TypeError: evaluate_js() missing 1 required positional argument: 'uid'

With this change the program already works.


r0x0r:
Thanks!

Issue #238, opened 2018-06-18
Javascript API stops working after reloading the page


mikahama:
### Specification
- Platform: mac
- Version: 2.0.3

### Description
On Mac, when I create the window, the API calls work as expected. If, however, I reload the page, by navigating to the same url again with webview.load_url, by running location.reload(); with javascript or just by right clicking on the page in debug mode and selecting reload, the API stops working. Inspector shows ReferenceError: Can't find variable: pywebview.

<img width="912" alt="captura de pantalla 2018-06-18 a las 19 17 26" src="https://user-images.githubusercontent.com/1543680/41548809-a4b5b930-732c-11e8-95c5-394b20d733f4.png">

r0x0r:
JS API and CSS changes currently do not persist after a page reload. This should definitely be remedied.  If you wish to give it a go, PRs are welcomed.

jrr8:
I found a partial workaround for this issue that will allow you to mimic a redirect. You may be able to modify this to work for a refresh as well.

Before I had a button that triggered a redirect from '/' to '/home', and the pywebview variable disappeared on the '/home' page, as you've noted.

The fix was instead of redirecting from '/' to 'home', I would AJAX fetch the content of '/home' from '/' and then call document.write() on the '/' page with the response from '/home'. This keeps the location at '/' but allows me to display the content of '/home' while still having access to the pywebview variable. Hope that helps.

r0x0r:
I pushed a macOS fix to the `fix-jsapi-reload` branch. 
@mikahama @jrr8 please test if the fix works for you.

Other platforms need to be addressed too.

r0x0r:
PR here 
https://github.com/r0x0r/pywebview/pull/248

r0x0r:
QT / GTK do not have this problem.
Windows preserves the API on url navigation, but not on refresh (eg. location.reload()). I cannot think of a solution for this, but at least it is documented here.

r0x0r:
Found a Windows solution. PR is here https://github.com/r0x0r/pywebview/pull/249

PR #248, opened 2018-08-28
#238: Fix disappearing JS API on reload

r0x0r:


jrr8:
I just checked the update on my machine and it appears to be working (i.e. pywebview.api persists on page reload)

PR #290, opened 2018-11-29
Set Window Size Functionality

aprowe:
Added set_window_size to api.
Added for all backends, tested on two of them

```
set_window_size(width, height, uid='master')
```

Tested Platforms:
- cocoa
- winforms

Untested platforms:
- win32
- gtk
- Qt

I wasn't able to test the win32, qtk, and Qt, so if someone wants to test those that would be great.

### DIFF
--- a/webview/cocoa.py
+++ b/webview/cocoa.py
@@ -348,6 +348,21 @@ class BrowserView:
         AppHelper.callAfter(toggle)
         self.is_fullscreen = not self.is_fullscreen
 
+    def set_window_size(self, width, height):
+        def _set_window_size():
+            frame = self.window.frame()
+
+            # Keep the top left of the window in the same place
+            frame.origin.y += frame.size.height
+            frame.origin.y -= height
+
+            frame.size.width = width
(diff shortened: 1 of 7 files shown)


webview/cocoa.py
```diff
@@ -348,6 +348,21 @@ def toggle():
         AppHelper.callAfter(toggle)
         self.is_fullscreen = not self.is_fullscreen
 
+    def set_window_size(self, width, height):
+        def _set_window_size():
+            frame = self.window.frame()
+
+            # Keep the top left of the window in the same place
+            frame.origin.y += frame.size.height;
+            frame.origin.y -= height;
```
r0x0r: semicolons

aprowe: Hah- I guess I just copy pasted from Obj-C code XD

aprowe: Fixed it though

aprowe:
Awesome, glad that works.

aprowe:
For most implementations It looks like a move-window function would be a small change

PR #294, opened 2018-12-09
[Cocoa] Upgrade to WKWebView

shivaprsdv:


webview/metadata.py
```diff
@@ -0,0 +1,25 @@
...
+                "retval": { "type": b"v" },
+                "arguments": {
+                    0: {"type": b"^v"},
+                    1: {"type": b"@"},
+                    2: {"type": b"@"}
+                }
+            },
+        }
+    }
+}
```
r0x0r: Could you move this code inside cocoa.py? It is Mac only and the name is confusing.

shivaprsdv: Of course. 👍 
I had thought we might need additional metadata for the delegate methods
also, that is why I added a separate file.

r0x0r:
Tests are failing. I will take a look.

r0x0r:
macOS tests fail as `webview_ready` is set in `applicationDidFinishLaunching_` which is called once per app. As I changed tests to a single process model, this obviously does not work. I investigated if there are any other delegates available and `windowDidLoad` of `NSWindowController` is an obvious  candidate. I gave it a quick test, but could not find an obvious way to get it working. @shivaprsdv any idea about this one?

shivaprsdv:
> macOS tests fail as `webview_ready` is set in `applicationDidFinishLaunching_` which is called once per app. As I changed tests to a single process model, this obviously does not work. I investigated if there are any other delegates available and `windowDidLoad` of `NSWindowController` is an obvious candidate. I gave it a quick test, but could not find an obvious way to get it working. @shivaprsdv any idea about this one?

`webview_ready` is set in the `show` method of each BrowserView instance,
except for the `master` instance, for which `show` will block. So the call in
`applicationDidFinishLaunching_` is just for the initial window. For all other
windows, the call in `show` should suffice. How does this fail to work?

We cannot use `windowDidLoad` as we don't use an `NSWindowController`
currently. I think that class is for loading windows from _.nib_ files (common
in Xcode projects).

r0x0r:
Tests are mess as always, but managed to get them pass on all platforms. I will merge this.

Issue #243, opened 2018-08-13
Edge support for Windows


r0x0r:
Microsoft has finally added support for hosting UWA controls in Windows Forms / WPF apps. Will be most appreciated if someone can test drive this. 

https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/xaml-host-controls

r0x0r:
The above docs are not clear on this, but apparently this requires .NET Core 3, which is not released yet.

shivaprsdv:
This is good news.. Winds of change 🙂

austinorr:
@shivaprsdv @r0x0r more [winds](https://blogs.windows.com/windowsexperience/2018/12/06/microsoft-edge-making-the-web-better-through-more-open-source-collaboration/) than expected? At first glance this seems like really good news.

r0x0r:
Yep. This news got me thinking if there is any point in implementing
edgehtml support at all when it arrives.

shivaprsdv:
Well well.. First GitHub, now Chromium, are you serious, Microsoft? 😄
This means we can get away without #15 as well. 🤷‍♂️

r0x0r:
Current progress is in the [edge-support](https://github.com/r0x0r/pywebview/tree/edge-support) branch. The basic functionality is there (including auto-detection of Edge support), but JS support is missing.

The main hurdle is that `InvokeScriptAsync` crashes the program for content loaded with `NavigateToString`. The bug is described [here](https://github.com/windows-toolkit/WindowsCommunityToolkit/issues/2165). It was apparently fixed, but the fix is not present in 5.1.1. Tried installing the preview version of WebView 6.0, but for some reason it did not work. 
There are a couple of solutions described in the issue. All involve using something else than `NavigateToString`. Anyhow the Edge is no different to other implementations - finding odd solutions for quirky bugs.

The architecture will need a major refactor at some point, but won't happen before pywebview 3 release.

r0x0r:
`WebViewControl` is buggy af.
In addition to the bug above  `file:// ` URLs are not supported and there is a bug related to crashing the program after closing a window via `window.close`. This makes the test runner crash.
Anyhow I made it work in a more or less adequate manner. I will not spend any more time polishing it at this point and will release 3.0 using the current version.

Known issues are documented [here](https://github.com/r0x0r/pywebview/blob/edge-support/docs/guide/renderer.md)

r0x0r:
[...]

Issue #272, opened 2018-10-22
New API


r0x0r:
This is going to be the first ever breaking change, so let's make this 3.0 once it is done. This matter was discussed [earlier](https://github.com/r0x0r/pywebview/issues/181). As this is quite a big change, I would like to get it implemented before other features to keep conflicts low.

The API would introduce a new function `webview.start(callback=None, parameters=(), multiprocessing=False)` that would start the UI loop and block program execution. The bare minimum hello world example would be then

```
import webview
window = webview.create_window("Test", "http://example.org")
webview.start()
```

Multiple window example would be 

```
import webview
window1 = webview.create_window("Test 1", "http://example.org")
window2 = webview.create_window("Test 2", "http://example.org")
webview.start()
```

All the webview functions related to the created window will be move to the Window object returned by `create_window`. For example Loading HTML would be 

```
import webview

def load_html(window):
  window.load_html("Hello world!")

window1 = webview.create_window("Test 1")
webview.start(load_html, window)
```

Moreover, I propose a new  function signature of `create_window`, so that all window options would be in a separate object and localization moved out of `create_window`. So that
```
webview.create_window(title, url=None, js_api=None, options={})

webview.localization = { ... } # default localization object
```

Specifying window option would be like this 

```
webview.create_window("Test", options={'resizable': False, 'fullscreen': True})
```

I am not sure what to do with the `confirm_quit` switch. Semantically it should be a global, not a window specific parameter, but I am not comfortable to moving it to `webview.config`.

It would not hurt to introduce `webview.windows` property that would contain all the opened windows.

@shivaprsdv Would you be able to give this a hand at some point?

mikaelho:
I like it!

Would this also enable calling start() from a subthread? I would really need to reserve the main thread for other stuff.

r0x0r:
GUI stuff must be run on the main thread. This is a requirement by every
underlying GUI framework. There is no way around it.

shivaprsdv:
> @shivaprsdv Would you be able to give this a hand at some point?
Certainly. We can start working on it. All your proposals are well formed. And let's
keep the discussions of #181 in mind. Let us see what we can turn up with.

r0x0r:
Glad to hear that! 
The roadmap would be #210 and #169 first and then I would like to start working on the new API. Maybe remove QT4 in process too.

HeavenVolkoff:
Are you considering to drop Python 2 support with this 3.0 release?
[...]

PR #306, opened 2019-01-29
WinForms with Edge WebView (WIP)

HeavenVolkoff:
Hello,

This is an initial implementation of the Edge WebView backend for pywebview running under Windows 10 (April 2018 Update and later)

The basic functionalities are implemented. e.g. Loading from URL, invoking javascript, JS_api, ...

However some are still missing:
 - File download
 - load_html

Also some are buggy: e.g. Alert dialog

This implementation is also lacking lots of tests. I only tested some basic websites and static html files on my machine, that is running Windows 10 version 1809 with Python 3.

This is my first time dealing with C#, .NET and Windows development. So, I would appreciate some guidance solving the remaining issues...

Solves https://github.com/r0x0r/pywebview/issues/243


r0x0r:
Thanks for the contribution. Much appreciated!
I will take a closer look late, but meanwhile a couple of words regarding the architecture. Considering this is WinForms based as well, it is better be done on top of  the existing `winforms.py` implementation with a healthy amount of conditions. I have done it this way in the CEF implementation [here](https://github.com/r0x0r/pywebview/blob/cef/webview/winforms.py). All the Edge initialisation could be done using inside a `_create_edge_browser` function. The best course of action would be probably to rebase your branch on top of the `cef` branch and go there. 
I will try to finish the CEF implementation soon to avoid conflicts.

HeavenVolkoff:
> Considering this is WinForms based as well, it is better be done on top of the existing `winforms.py` implementation

I constructed this on top of `winforms.py` by replacing the Trident specific code with what was needed to setup the Edge WebView. However I agree that it would be better to leave all the generic WinForms initializations in one place for easier future maintainability. I will start working on it as soon as I get some more spare time from work.

> with a healthy amount of conditions. I have done it this way in the CEF implementation [here](https://github.com/r0x0r/pywebview/blob/cef/webview/winforms.py). All the Edge initialisation could be done using inside a `_create_edge_browser` function. The best course of action would be probably to rebase your branch on top of the `cef` branch and go there.

If you don't mind me making a suggestion : I checked your CEF implementation code, and IMHO, I think a solution using a base class with all the WinForms initialization code, and subclasses with the specific code for each webview approach would be more fitting. I could also help out with the CEF implementation, if needed.

r0x0r:
Class inheritance sounds good to me, as long as it is not overly complicated just for the sake of it. The CEF implementation is platform agnostic (even though it works only on Windows), so the subclassed version of winforms would be just some form of a glue component between `winforms.py` and `cef.py`. I also would like to keep all the winforms related stuff in `winforms.py` for now.

The Edge should the default renderer for compatible systems Windows 10 >= version 1803 and .NET Framework >= 4.6.2. As long as the requirements are met, it should be selected by default. I think there is no need for a GUI switch.

r0x0r:
There is a major [refactoring](https://github.com/r0x0r/pywebview/issues/272) planned for the version 3.0, which will introduce breaking API changes and will probably result in consolidation of different implementations. So I would not put much effort into getting the current architecture right at this point.

By `winforms stuff` I mean everything WinForms related, incuding Trident, Edge and CEF glue code.

Here is how you programmatically get a .NET version.
https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?redirectedfrom=MSDN&view=netframework-4.7.2#System_Environment_Version

r0x0r:
@HeavenVolkoff 
Here is the CEF implementation https://github.com/r0x0r/pywebview/pull/310

HeavenVolkoff:
Nice work,
As soon as I get some more spare time I will take a look on your changes and see how to better adapt my code to fit everything together.

r0x0r:
@HeavenVolkoff 
API refactoring is more or less complete. I will start looking into incorporating this into the code. I will go with the same route as with CEF, unless you have a better architecture in mind.

HeavenVolkoff:
Hey man,
Sorry for the delay answering. I really want to keep working on this, and I have some neat solutions to implement and merge into the new architecture. However, I am very short on time (due to work and college). So I will only be able to continue this around July. If you want you can merge this as is, but the caveats I mentioned in the first post are still there. Alternatively, we could hold this for a future minor release. Up to you.

r0x0r:
@HeavenVolkoff 
I incorporated your changes in the [edge-support](https://github.com/r0x0r/pywebview/tree/edge-support) branch using the new architecture. 
Anyway let's continue discussion [here](https://github.com/r0x0r/pywebview/issues/243). I am closing this PR.

PR #332, opened 2019-04-25
Show the default window control buttons.

Fizzadar:
It seems pretty user-unfriendly to remove these. This matches (most?) other frameless MacOS apps (Slack, Standard Notes, etc).

Could also make this an option (`window_controls=True`)?


r0x0r:
@Fizzadar 
I am working on pywebview 3.0's API at the moment. Will get back to this after I finish with the new API.

r0x0r:
@Fizzadar I still have no opinion about this matter. As nobody else does either, let's go with your proposal.
I merged your changes into master manually. Closing this PR now.

PR #337, opened 2019-06-08
Use os.remove rather than shutil.rmtree to remove error.log on exit

cormacc:
shutil.rmtree is only supposed to take a directory argument according
to docs. Was resulting in an exception on window shutdown.


cormacc:
Observed the error on Windows 10 64 btw.

PR #349, opened 2019-07-17
Don't escape line breaks in result of js_bridge_call

kvasserman:
Escaping line breaks in results of the python calls causes problems when result object contains strings with line breaks. What you get on js side is strings with '\n' as literal strings (a slash followed by letter n) instead of the special new-line character. This basically mangles the data being passed from python to js in an unrecoverable manner: js code has no way of knowing if it was meant to be the literal text or a special character.

json.dumps() already escapes everything that needs to be escaped and creates a valid string to be passed to the js side. JSON.parse() correctly reverses the data to it's original state. I don't think there is a need to mess with special characters here.

### DIFF
--- a/webview/js/api.py
+++ b/webview/js/api.py
@@ -46,7 +46,7 @@ window.pywebview = {
             if (returnObj.isSet) {
                 returnObj.isSet = false;
                 try {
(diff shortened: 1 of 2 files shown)


r0x0r:
Thank for the PR. To be honest I have no longer any idea where and when escapes are needed, but this change makes sense to me.

Issue #357, opened 2019-07-31
raise Permission denied when create_window with params html


zwgshr:
### Specification
- pywebview version: 3.0.1
- platform / version:Windows 10 1903 Build 18362.267

### Description  

Here the Python code:
``` python
window = webview.create_window(title='test', html='loading...', js_api=api)
webview.start(change_url, window)
```

the simple example script fails with the message:
```
未经处理的异常:  Python.Runtime.PythonException: PermissionError : [Errno 13] Permission denied: 'C:\\Program Files\\Python37\\b98d1ad1f1094333bd2064d1715ced32.html'
   在 Python.Runtime.Dispatcher.Dispatch(ArrayList args)
   在 System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   在 System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   在 System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   在 System.Threading.ThreadHelper.ThreadStart()
```

How I fixed it

``` python
# /webview/platforms/winforms.py
        def load_html(self, html, base_uri):
            file_name = '%s.html' % uuid4().hex
--            self.temp_html = os.path.join(WinForms.Application.StartupPath, file_name)
++            self.temp_html = os.path.join(WinForms.Application.UserAppDataPath, file_name)

```

### Practicalities

- [x] I am willing to work on this issue myself.

- [ ] I am prepared to support this issue financially.

r0x0r:
Have you tested that it works with EdgeHTML? The thing is this stupid
workaround requires that html file is in the same folder as executable. I
could not get it work from another location.

I don't have access to a Windows computer at the moment, so cannot test it
myself.

zwgshr:
![image](https://user-images.githubusercontent.com/10608334/62209430-25c6da00-b3cc-11e9-90f3-19c5d46fd65d.png)

It works with EdgeHTML, UserAppDataPath is better than StartupPath.

r0x0r:
Fix is deployed to 3.0.2

PR #358, opened 2019-07-31
Change winforms tmp_html path

zwgshr:
#357


r0x0r:
@zwgshr It is not working for me. With WinForms.Application.UserAppDataPath, HTML is never loaded.  According to the EdgeHTML documentation, NavigateToLocal looks for a file in the same directory as the executable

r0x0r:
I released 3.0.2, in which http server is used for local urls and load html. I am closing this PR now.

PR #365, opened 2019-08-06
Update guilib.py

adbenitez:
if `forced_gui == 'gtk'` then  `forced_gui != 'qt'` so it is redundant here, we can remove it.


r0x0r:
Good catch. I got rid of negation and switch if else.

Issue #351, opened 2019-07-17
Option to start window centered and disabling scrollbars


mariolatiffathy:
### Specification
- pywebview version: 3.0
- platform / version: Windows 8.1 64-bits

### Description  
  
  Describe your issue

The PyWebView window starts with a random position on the screen, I'd like to start it centered but it seems PyWebView doesn't have such feature. It'd be appreciated if this could be added.
I've seen a similar issue here: https://github.com/r0x0r/pywebview/issues/158 , but the owner of the issue closed it immediately.
_NOTE_: It will be even better if we can set the X and Y position, so we can do like `position=(0, 0)` to center the window, instead of `position="centered"`

Also there should be an option to disable scrollbars as well, I currently disable them using CSS and JavaScript in HTML itself, but it'd be better if this can be disabled in the window through an option like `scrolling=False`.

r0x0r:
My take on this is that the window should be centered by default with an option to set a custom position with  the `position=(x,y)` parameter.

I see no value in the `scrolling` parameter,  as it would be set with CSS anyway.

r0x0r:
PR is open https://github.com/r0x0r/pywebview/pull/402
macOS logic does not work with multiple screens. I will leave it as is at this point.

PR #367, opened 2019-08-06
[WIP] add move function

adbenitez:
partially tackles #351 adding a `window.set_window_position(x, y)` implemented **only for gtk and qt** platforms, sorry, I am not a Mac or Windows and have no experience with `cocoa` or `winforms`.
Leaving here as base for someone else to implement the rest of platforms :)

### DIFF
--- a/webview/platforms/gtk.py
+++ b/webview/platforms/gtk.py
@@ -245,6 +245,9 @@ class BrowserView:
     def set_window_size(self, width, height):
         self.window.resize(width, height)
 
+    def move(self, x, y):
+        self.window.move(x, y)
+
     def create_file_dialog(self, dialog_type, directory, allow_multiple, save_filename, file_types):
         if dialog_type == FOLDER_DIALOG:
             gtk_dialog_type = gtk.FileChooserAction.SELECT_FOLDER
@@ -398,6 +401,12 @@ def set_window_size(width, height, uid):
     def _set_window_size():
         BrowserView.instances[uid].set_window_size(width,height)
     glib.idle_add(_set_window_size)
+
+
+def move(x, y, uid):
+    def _set_window_position():
+        BrowserView.instances[uid].set_window_position(x, y)
+    glib.idle_add(_set_window_position)
 
 
 def get_current_url(uid):
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -427,6 +427,21 @@ class BrowserView:
 
         AppHelper.callAfter(_set_window_size)
 
+    def move(self, x, y):
+        def _move():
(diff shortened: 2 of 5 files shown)


adbenitez:
oops! sorry :( I for got I have autopep8 to format automatically on save in my Emacs, feel free to modify this PR if you don't want to change the formatting

r0x0r:
Thanks for tackling this! Could you resubmit the PR with autoformat off? 
Another thing, a more apt name for this function could be `window.move(x,y)`

adbenitez:
I will go offline for today, I will not have internet until tomorrow, will do it then, sorry :(

adbenitez:
@r0x0r 
I blindly implemented the move function in winforms.py, it *should* work ;)
I noticed here that the code is using a int for the flags of `setWindowPos` I think for better documentation would be better to define some global constants with the flags and use them as I did in the move function, also the `set_window_size` function may use other flags like `SWP_NOMOVE` to just pass the width and height and ignore the x and y position since this is not a move operation, so you don't need to use `self.Location.X` etc.
(I don't know anything about this windll thing I am just guessing so sorry if I am making a dummy suggestion ;)

adbenitez:
about `qt.py`  tell me if you are ok with the internal names `window_move_trigger` and `on_window_move`?

adbenitez:
added the implementation for `cocoa.py` this really needs a review I am not really sure of what I have done there, seems like y coordinates are flipped in cocoa

acbetter:
> added the implementation for `cocoa.py` this really needs a review I am not really sure of what I have done there, seems like y coordinates are flipped in cocoa

`cocoa.py` works well with this test file on my macOS.

```python
# test_move_window.py

import time
import webview
from .util import run_test


def test_move_window():
    window = webview.create_window("Set Window Size Test", "https://www.example.org")
    run_test(webview, window, move_window)


def move_window(window):
    window.move(50, 500)
    time.sleep(1)
    window.move(500, 50)
    time.sleep(1)
    window.move(500, 500)
    time.sleep(1)
    window.move(50, 50)
    time.sleep(1)
```

webview/platforms/gtk.py
```diff
@@ -400,6 +403,12 @@ def _set_window_size():
     glib.idle_add(_set_window_size)
 
 
+def move(x, y, uid):
+    def _set_window_position():
+        BrowserView.instances[uid].set_window_position(x, y)
```
r0x0r: set_window_position instead of move

r0x0r:
I have made some fixes and moved the PR to `move-window` branch. Some things left to do:
- Window centering on GTK and Windows.
- Cocoa screen coordinate calculation in a multiscreen setup.

PR #392, opened 2019-10-09
Bump PyTest to Fix CI

rotu:
Bump pytest version to avoid crash.


rotu:
Did not intend to close this. i blame my browser

PR #395, opened 2019-10-10
Add missing dependency for GTK

rotu:
Without this package, loading gtk fails with:
[pywebview] GTK cannot be loaded
  File "/home/dan/Documents/pywebview_ws/pywebview/webview/guilib.py", line 16, in import_gtk
    import webview.platforms.gtk as guilib
  File "/home/dan/Documents/pywebview_ws/pywebview/webview/platforms/gtk.py", line 25, in <module>
    import gi
ModuleNotFoundError: No module named 'gi'

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,7 @@ from setuptools import setup
 extras_require = {
     'qt': ['PyQt5', 'pyqtwebengine'],
     'cef': ['cefpython3'],
+    'gtk': ['PyGObject'],
 }
 
 install_requires = [


r0x0r:
Does pygobject take care of all the needed dependencies? I have installed GTK dependencies using apt ie
https://pywebview.flowrl.com/guide/installation.html#linux

rotu:
It does not. If Gtk is not installed, this will still crash at the line `gi.require_version('Gtk', '3.0')`. This just declares the python dependencies so you can pip install pywebview to be installed into a clean virtualenv if you already have GTK installed system-wide.

Issue #220, opened 2018-05-08
Use Qt GUI per default on Qt-based systems


flying-sheep:
Users on e.g. Plasma or LXQt want this to use Qt without having to define a environment variable.

shivaprsdv:
You don't really need the environment variable, you could put:
```py
webview.config.use_qt = True
```
at the top of your script, as in the `qt_test.py` example.

Anyway if PyGObject is missing, as it is likely on Qt-based systems, pywebview
will default to PyQt.

flying-sheep:
i respectfully disagree in both points:

- on many systems, quite some stuff is installed, which might include pygobject as some dependency. i have it, and i try to keep GTK out as much as feasible
- I know that it’s easy to do progammatically for individual pieces of software, but i want the default to be sensible. Qt on a Qt system, GTK on a GTK system

sadly the `desktop` package is very old, else we could have used that.

the only thing i know out of the top of my head is that if `KDE_FULL_SESSION` is defined, it should be treated as `USE_QT`.

shivaprsdv:
Yes, that makes sense. I'm a school kid in the Linux user land, so thanks
for the tip. Your suggestion is definitely feasible.

r0x0r:
How would you detect which one to default to?

r0x0r:
Done

adbenitez:
I think this most to be analyzed more carefully, what about if now comes a programmer saying that he don't want to use QT if GTK is available??? 
With this approach we can't know if `use_qt` was set by the programmer or by pywebview, also giving an error in a virtual environment with GTK but without QT.
A better approach could be simply check for QT before for GTK in this "Qt-based" systems (with a list of 2 functions and changing the order based on `KDE_FULL_SESSION`, as an example), this way we can know if the programmer wants explicitly  to use QT (and throwing an error if is missing) or just try to use GTK if QT is missing.
This is just a recommendation so feel free to just ignore it :)

r0x0r:
This is a valid point.
QT used to be default, but I switched to GTK at some point, as I felt it
was more popular (I blame Ubuntu). I would not like to make a switch back,
unless somebody convinces me on the QT superiority :)
A crude solution would be to introduce a USE_GTK variable that would force
GTK on QT-based systems, while keeping the current preference order.

flying-sheep:
sure, optimally:

- the default depends on your DE, with some smart detection going on
[...]

PR #396, opened 2019-10-10
Changed platform-loading logic

rotu:
1. Use importlib.import_module instead of nearly identical import statements.
2. Allow passing a list instead of just one string, to fully customize the order that we try guis.
3. ~~Remove confusing logic to ignore gtk if envvar KDE_FULL_SESSION is set.~~

### DIFF
--- a/webview/__init__.py
+++ b/webview/__init__.py
@@ -77,6 +77,7 @@ def start(func=None, args=None, localization={}, gui=None, debug=False, http_ser
         raise WebViewException('You must create a window first before calling this function.')
 
     guilib = initialize(gui)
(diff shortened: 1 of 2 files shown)


r0x0r:
KDE_FULL_SESSION is checked in order to prefer QT over GTK on QT systems. There is a rather long discussion covering this issue [here](https://github.com/r0x0r/pywebview/issues/220)

rotu:
Reverted that part. I don't understand the over/under on QT vs GTK. Is there a reason to not just make QT take priority over GTK by default in all cases (and just use GTK if QT is unavailable)?

PR #416, opened 2019-11-01
[WIP] Get size + position

Fizzadar:
WIP - wanted to get your thoughts before attempting to implement for QT/etc!


r0x0r:
Cool, please go ahead.

I am not too keen on the names `get_position` and `get_size`, as with time it will be a bunch of `get_` methods. I cannot think of better names, though.
Maybe do it with properties?  E.g. `window.width`, `window.height`, `window.x`, and `window.y`. That would present a problem with setters, ie. should those be writeable as well. If so, then these properties would duplicate the functionality of `window.resize` and `window.move`. 
Thoughts are welcomed.

Fizzadar:
@r0x0r agreed - the two get's are no good; I see two options:

+ `get_frame` (or similar) that returns both the x/y/width/height (as all require the same underlying call)
+ read-only window properties - each would need to make a different call to get the frame (in cocoa) which might not be efficient - but I do like the idea of a `window.width` and so on, it's a much nicer API for the end-user

What do you think? In the meantime will investigate implementing this for QT.

r0x0r:
Let's go with read-only properties for now: x, y, width, height. If a need for a frame arises, it can be implemented later.

Fizzadar:
bump. Hoping to get time to finish this up this week!

Fizzadar:
@r0x0r have just pushed a re-implementation using properties - let me know what you think!

r0x0r:
@Fizzadar looks good to me.
Would you be able to provide implementations for other platforms?

Fizzadar:
@r0x0r yes I'm going to look into the other platforms over the holidays; only issue might be Winforms as I've no proper Windows testing environment.

Fizzadar:
@r0x0r much delayed but have just pushed working implementations for QT+GTK (both tested on Mac).

r0x0r:
There is an issue with properties, namely they must be wrapped with the `@_shown_call` decorator. Now calling any of the properties immediately after `webview.start()` results in an error.

r0x0r:
There is a problem with retrieving window coordinates on Cocoa too.
If you set x=100, y=200 in `create_window` and retrieve them right after it will return x=100.0, y=100.0.

r0x0r:
I have pushed my changes to the `Fizzadar-get-size-position` branch. Please update the PR with it.

There is a problem with retrieving correct window width on Cocoa as well, as demonstrated by `examples/resize.py`.

r0x0r:
I have added Windows implementation, documentation and tests. The changes are in the `Fizzadar-get-size-position` branch.

Fizzadar:
Thank you; have merged in your changes from the `Fizzadar-get-size-position` branch.

PR #436, opened 2019-11-26
Implement JS API error object

r0x0r:


webview/util.py
```diff
@@ -101,7 +101,12 @@ def _call():
             result = json.dumps(result).replace('\\', '\\\\').replace('\'', '\\\'')
             code = 'window.pywebview._returnValues["{0}"]["{1}"] = {{value: \'{2}\'}}'.format(func_name, value_id, result)
         except Exception as e:
-            result = json.dumps(traceback.format_exc()).replace('\\', '\\\\').replace('\'', '\\\'')
+            error = {
+                'message': str(e),
+                'name': type(e).__name__,
```
RyannnnnnR: A small knit-pick but I think `name` should be renamed to `type`

r0x0r:
name is a property of the JS Error object. toString() uses it too, so it is better to stay it is.

PR #454, opened 2020-01-10
Added Chinese readme documentation

linklinco:
Added Chinese readme documentation

### DIFF
--- a/README_zh.md
+++ b/README_zh.md
@@ -7,21 +7,21 @@ https://pywebview.flowrl.com
 
 [English](README.md) | [简体中文](README_zh.md)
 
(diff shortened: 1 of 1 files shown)


r0x0r:
I don't understand a word of it, but is there a need for Chinese readme?

linklinco:
I think this is useful to help this project internationalize，and I plan to translate the entire document。:-)

r0x0r:
I welcome this effort, however I have got concerns about maintainability.
As I am unable to keep the Chinese documentation up-to-date, it should be
done by someone else. Would be able to commit to this task?

With documentation in different languages there is also a chance of being
differences between versions. So some kind of versioning system should be
defined to note whether documentation is synced. I recently put a pywebview
version number on the documentation front page, maybe that will be enough.

>

PR #469, opened 2020-02-11
Support python2.7: change getfullargspec() to getargspec()

zvibazak:
In order to support python2.7 change getfullargspec() to getargspec().

### DIFF
--- a/webview/util.py
+++ b/webview/util.py
@@ -75,7 +75,7 @@ def parse_file_type(file_type):
 
 def parse_api_js(window, platform, uid=''):
     def get_args(f):
-        return list(inspect.getfullargspec(f).args)
+        return list(inspect.getargspec(f).args)
 
     def generate_func():
         if window._js_api:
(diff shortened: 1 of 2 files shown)


r0x0r:
I am phasing out support for Python2. Is there a use case for keeping this Python2 compatible?

zvibazak:
Yes we still use python 2   :)
I belive that there are another developers that the migration is too hard to them.

r0x0r:
While I have no problems accepting this PR, I am not able to support Python2. Meaning that future changes might break Python2 compatibility once again. In a nutshell it is going to be somebody else’s responsibility to verify that pywebview works under Python 2.

zvibazak:
Ok sounds great. I’ll push the fix I see.

r0x0r:
getargspec has been deprecated since Python 3.0. Please introduce a condition to use getfullargspec if running under Python 3.

zvibazak:
Done.

PR #478, opened 2020-02-26
added topmost and toggle_topmost, for Windows.

### LINKED ISSUE #476: Always-in-front / Always-on-top window, on Windows
### Specification
- pywebview version: 3.2
- platform / version: Windows 10

### Description  
  
  Not an issue so much as a question. Is it possible to keep a window in front of all other windows, for example for a small control panel in the corner? tkinter has something like `.wm_attributes("-topmost", 1)` to keep a window in front. Maybe Python has a way to do this via some Windows api, even if it's not part of pywebview api.
  
### Practicalities

- YES/NO I am willing to work on this issue myself.

- YES/NO I am prepared to support this issue financially.

N/A just asking a question.

tscizzlebg:
Closes #476 

Added examples, updated docs, added a trivial test that things don't error out.

Tested on my Windows machine that starting a webview window with topmost=True does in fact keep that window above others.

### DIFF
--- a/webview/__init__.py
+++ b/webview/__init__.py
@@ -103,7 +103,8 @@ def start(func=None, args=None, localization={}, gui=None, debug=False, http_ser
 
 def create_window(title, url=None, html=None, js_api=None, width=800, height=600, x=None, y=None,
                   resizable=True, fullscreen=False, min_size=(200, 100), hidden=False, frameless=False,
-                  minimized=False, confirm_close=False, background_color='#FFFFFF', text_select=False):
+                  minimized=False, topmost=False, confirm_close=False, background_color='#FFFFFF',
+                  text_select=False):
     """
     Create a web view window using a native GUI. The execution blocks after this function is invoked, so other
     program logic must be executed in a separate thread.
@@ -117,6 +118,7 @@ def create_window(title, url=None, html=None, js_api=None, width=800, height=600
     :param hidden: Whether the window should be hidden.
     :param frameless: Whether the window should have a frame.
     :param minimized: Display window minimized
+    :param topmost: Keep window above other windows (required OS: Windows)
     :param confirm_close: Display a window close confirmation dialog. Default is False
     :param background_color: Background color as a hex string that is displayed before the content of webview is loaded. Default is white.
     :param text_select: Allow text selection on page. Default is False.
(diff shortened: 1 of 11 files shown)


tscizzlebg:
#476

webview/__init__.py
```diff
@@ -103,7 +103,8 @@ def _create_children(other_windows):
 
 def create_window(title, url=None, html=None, js_api=None, width=800, height=600, x=None, y=None,
                   resizable=True, fullscreen=False, min_size=(200, 100), hidden=False, frameless=False,
-                  minimized=False, confirm_close=False, background_color='#FFFFFF', text_select=False):
+                  minimized=False, topmost=False, confirm_close=False, background_color='#FFFFFF',
```
r0x0r: I prefer the param name `on_top`. Topmost sounds too Windows-specific to me.

webview/window.py
```diff
@@ -248,6 +249,13 @@ def toggle_fullscreen(self):
         """
         self.gui.toggle_fullscreen(self.uid)
 
+    @_shown_call
+    def toggle_topmost(self):
+        """
+        Toggle topmost mode
+        """
+        self.gui.toggle_topmost(self.uid)
+
```
r0x0r: Though there is a `toggle_fullscreen`, I am not convinced about introducing more `toggle_` or `set_` functions. We have had a similar dicussion with x, y, width and height attributes and it was decided to implemented those as class properties. Maybe this one could be implemented in a similar fashion, ie `window.on_top = True`.
Opinions?

webview/platforms/winforms.py
```diff
@@ -491,6 +495,18 @@ def _toggle():
             else:
                 _toggle()
 
+        def toggle_topmost(self):
+            def _toggle():
+                if not self.is_topmost:
+                    self.TopMost = True
+                else:
+                    self.TopMost = False
```
r0x0r: Can be simplified as `self.TopMost  = not self.is_topmost`

tscizzlebg:
Heads up, that unrelated to this feature, my team moved forward with a different package instead, so I'm not going to continue on this. Thanks for the responsiveness.

N0ich:
Any chance this feature will make it?
I can check the changes if needed

r0x0r:
Sure. There is a PR opened with my comments. As soon as somebody will make fixes, I will merge it.

PR #491, opened 2020-03-24
on_top - toggle_on_top (#478 followup) for windows

N0ich:
https://github.com/r0x0r/pywebview/pull/478

I corrected the requested changes. 
As far as i'm aware, the initial PR was also missing the `windll.user32.SetWindowPos` required to set the zorder value. 
I've kept the `toggle_on_top()` logic because i think it's better to be able to toggle it than having to relaunch if you want to change the behaviour, but i'm open to passing it as a propriety if you prefer.
(Honestly it fits better in my project to be able to toggle it)
I've added it and tests are okay.
Please let me know if i've made a mistake or something. 
Thanks for reading


r0x0r:
Good job!
I agree about the toggle_on_top() logic, but I am not fond of the name. Granted there is `toggle_fullscreen`, but it was introduced early on and I would not like another `toggle_*` or `set_*` method.  In my opinion properties are the most elegant solution, ie `window.on_top = False`. Window coordinates and position are currently implemented as read-only properties. Maybe this one can be implemented as a read-write property.

N0ich:
I'll look into it before the end of the week, no idea how to implement that. Thanks for the reply, will update.

N0ich:
Alright that was interesting
on_top is a property of window and the browser
This is easily applicable to fullscreen, i can take a look if you want

r0x0r:
Thanks for making the changes. I merged this into the `ontop` branch  and will implement other platforms when I get a chance.

Issue #498, opened 2020-04-11
Add support for custom user-agents


tognee:
### Description  
For the app I'm working on I need to open a new window and load a site with a specific user-agent and right now that's not possible.
I added support for it in gtk (as it's what I'm using now). (#499)

### Practicalities
I am willing to work on this issue myself.

r0x0r:
Some implementation links

WinForms / MSHTML https://stackoverflow.com/questions/937573/changing-the-user-agent-of-the-webbrowser-control
WinForms / EdgeHTML https://www.pedrolamas.com/2017/03/21/setting-a-custom-user-agent-in-the-uwp-webview-control/ (seems tricky)
CEF: https://github.com/cztomczak/cefpython/blob/master/docs/Tutorial.md#change-user-agent-string

QT: https://doc.qt.io/qt-5/qwebengineprofile.html#setHttpUserAgent

Cocoa: https://stackoverflow.com/questions/28109668/osx-10-10-wkwebview-set-useragent-swift

tognee:
gtk, qt, cocoa, cef and mshtml are now implemented in my pull request and tested.

The only one left is EdgeHTML and as you said it seems tricky. I don't think I can implement that one by myself as I'm not that fluent with c#

r0x0r:
@tognee I have implemented a global user agent via `start(user_agent='')`. The settings approach is there too, it can be used as a way to set a user agent individually per window (would not work in CEF though). PR is here if you want to take a look https://github.com/r0x0r/pywebview/pull/509

tognee:
Looks fine to me!
I think it can be merged without any problems.

Should I wait for the merge before closing this issue?

PR #499, opened 2020-04-11
Added user-agent support

tognee:
Added user-agent support for gtk, in reference to #498


r0x0r:
I am not fond of passing user_agent to a window. 
A parameter to `webview.start(user_agent='')` similar to gui would me a more apt choice imo, even though it would make user agent global for the application.

tognee:
I've looked into other renderers and it seems like not all have this option to change the user-agent.

For my usecase I need to open a window, get some data from a site that's only obtainable with a specific user-agent and then close the window and send the data to the frontend.
Using a global user-agent should work as well, but then the entire app uses that user-agent.

r0x0r:
Personally I see no problem in having a global user-agent. At least I cannot think of any use-case where it would be beneficial to have a unique user-agent for each window. 
Another solution could be setting a user agent in a similar war as Chrome settings are passed to CEF, ie https://github.com/r0x0r/pywebview/blob/master/examples/cef.py

tognee:
I've now removed the user_agent reference from the window __init__ and implemented the settings method already used by cef.
Did a test on gtk, qt, cocoa and cef and it works.

Now winforms is the only one missing. Need to make a VM to do MSHTML and then I will look into EdgeHTML

tognee:
Ok, so I've tested MSHTML on a Win7 VM and it works as well.

Now only EdgeHTML is left. I'm not good with c# so I think I'll need help with that one

r0x0r:
Good job. Would you mind adding an example too?
I would not stress about EdgeHTML, it can be marked as unsupported for now. Microsoft is promising Chromium based WebView2 sometime soon, so it looks like EdgeHTML will become obsolete in any case.

r0x0r:
Let's user_agent parameter to the start function to make it more accessible. I will continue from that.
Thanks for the contribution!

Issue #484, opened 2020-03-13
cef hangs while resizing in certain conditions


MarkusFF:
### Specification
- pywebview version: 3.2
- platform / version: Win7 Pro SP1
- cefpython3 version 66.0

### Description  
  
  Observed reproducible hang during use of cef gui, if triggering some kind of update while resizing the window. Minimally this code usually (but not 100% of the time) hangs on my machine:

```python
import webview
window=webview.create_window('Hello world', html='<h1>hello</h1>')
def reload_and_resize():
    window.evaluate_js("location.reload()")
    window.resize(1000, 420)
webview.start(reload_and_resize, gui="cef")
```

or while manually resizing something more like this:

```python
import webview,time
window=webview.create_window('Hello world', html='<h1>hello</h1>')
def keep_reloading():
    while True:
        time.sleep(1)
        window.evaluate_js("location.reload()")
webview.start(keep_reloading, gui="cef")
```

This may be a cefpython issue.

MarkusFF:
Could be this: https://github.com/cztomczak/cefpython/issues/464

r0x0r:
I have reproduced it and it looks like a race condition issue to me. At least adding a debug print statement somewhere in an execution chain seem to eliminate the problem. I will look into it.

ikhmyz:
> I have reproduced it and it looks like a race condition issue to me. At least adding a debug print statement somewhere in an execution chain seem to eliminate the problem. I will look into it.

Is there any news?

r0x0r:
I gave it a try, but did not make any progress. The problem is that as soon as I add any debugging statements, the problem is gone. For example

This hangs
```
import webview
window=webview.create_window('Hello world', html='<h1>hello</h1>')
def reload_and_resize():
    window.evaluate_js("location.reload()")
    window.resize(1000, 420)
webview.start(reload_and_resize, gui="cef")
```

This does not 
```
import webview
window=webview.create_window('Hello world', html='<h1>hello</h1>')
def reload_and_resize():
    window.evaluate_js("location.reload()")
    print('w00t')
    window.resize(1000, 420)
webview.start(reload_and_resize, gui="cef")
```
Similarly adding print statements elsewhere in the CEF code eliminates the problem. Breakpoints set in sub threads are not invoked either. A dumb solution would be add a short sleep noop.

The second example in the original post did not result in hanging on my machine.

ikhmyz:
[...]

PR #504, opened 2020-04-23
Fix CEF win resize hangs. https://github.com/r0x0r/pywebview/issues/484

ikhmyz:
Temporary solution until cefpython release 66.1 version (after that `Browser.SetBounds` method will be available for Win and we should use it to resize window)


docs/guide/installation.md
```diff
@@ -31,6 +31,12 @@ pip install pywebview[cef]
 pip install cefpython3
 ```
 
+pywin32
+
+``` bash
+pip install pypiwin32
+```
+
```
r0x0r, line 39: pywin32 is not needed. You can use `from ctypes import windll`

webview/platforms/cef.py
```diff
@@ -99,13 +103,18 @@ def initialize(self):
...
+        sleep(0.1)  # wait for window.pywebview to load
         self.initialized = True
         self.loaded.set()
 
     def close(self):
         self.browser.CloseBrowser(True)
 
+    def resize(self, width, height):
+        win32gui.SetWindowPos(self.inner_hwnd, win32con.NULL, 0, 0, width - 16, height - 38,
+                              win32con.SWP_NOZORDER | win32con.SWP_NOMOVE | win32con.SWP_NOACTIVATE)
```
r0x0r, line 115: `windll.user32.SetWindowPos`

ikhmyz: It's strange but using ctypes.windll I got hangs when I randomly resize window by dragging it (but with much less chance than using cef.WindowUtils.OnSize). In the same case I got no hangs using win32gui during much longer test. But it still could be random....

NiceneNerd:
What's the status of this? I have a project that's kind of waiting for this fix.

r0x0r:
Merged. I have not tested it personally.

PR #513, opened 2020-05-09
Internal Server Revamp

### LINKED ISSUE #483: More advanced local server?
### Specification
- pywebview version: 3.2
- platform / version: Linux+gtk

### Description  
  
The built-in local server is fine for really basic stuff, but:

* You can't use `importlib.resources` with it
* You can't "mount" several directories
* You can't fallback to something else

Since `wsgiref` has been part of the standard library for a while, doing some kind of basic wsgi routing middleware is doable without additional dependencies.

The basic concept is that you "mount" various sources at directory prefixes, and it'll select the most appropriate one. Sources include:

* A filesystem directory
* A python package (data loaded via `importlib.resources`)
* Another WSGI app

I think these three represent the most common uses while providing progressive enhancement. (It's very easy to "scale up" with this system as an application becomes more complex, without hitting any walls.)

The routing algorithm is simple: Use the longest matching directory prefix. "Directory prefix" means that `/foo/bar` will match `/foo/bar`, `/foo/bar/` and `/foo/bar/spam` but not `/foo/bar-spam`. Longest means that you can mount `/` and `/static` and it'll work as you expect.

I think package data (`importlib.resources`) should be the preferred way of loading application assets (although I haven't seen how it works with py2exe and friends). Directories are useful for serving user data, generated files, cache directories, etc. WSGI apps just allow open-ended anything.
  
### Practicalities

- YES I am willing to work on this issue myself.

- NO I am prepared to support this issue financially.

astronouth7303:
Rewrites the internal server to be wsgi-based, dramatically improving backend support.

This also includes features for handling static files and URL routing.

Closes #483

This uses [`wsgiref.simple_server`](https://docs.python.org/3/library/wsgiref.html#module-wsgiref.simple_server), which is based on `http.server`. While this isn't secure or terribly performant, it is simple and is probably good enough for single-user local stuff (ie this use case).

The interface I'm going with is that you may pass a WSGI app (which includes the included static and routing options) any place the public API expects a URL, and pywebview will manage it. (Of course, the user can elect to manage the server themselves and pass a URL to pywebview.)

This means that the user isn't required to mess with ports and threading and stuff. So using an existing wsgi-compatible application could be as simple as:

```python
import webview
import myapp.wsgi

webview.create_window('test', myapp.wsgi.app)
webview.start()
```

### DIFF
--- a/webview/window.py
+++ b/webview/window.py
@@ -75,7 +75,6 @@ class Window:
         self.shown = Event()
 
         self.gui = None
-        self._httpd = None
         self._is_http_server = False
 
     def _initialize(self, gui, multiprocessing, http_server):
@@ -83,9 +82,9 @@ class Window:
         self.loaded._initialize(multiprocessing)
         self.shown._initialize(multiprocessing)
         self._is_http_server = http_server
-
+        
         if http_server and self.url and self.url.startswith('file://'):
-            self.url, self._httpd = start_server(self.url)
+            self.url = start_server(self.url)
 
     @property
     def width(self):
@@ -152,14 +151,10 @@ class Window:
         :param url: url to load
         :param uid: uid of the target instance
         """
-        if self._httpd:
-            self._httpd.shutdown()
-            self._httpd = None
-
         url = transform_url(url)
 
         if (self._is_http_server or self.gui.renderer == 'edgehtml') and url.startswith('file://'):
-            url, self._httpd = start_server(url)
+            url = start_server(url)
 
         self.gui.load_url(url, self.uid)
 
@@ -172,9 +167,6 @@ class Window:
         :param base_uri: Base URI for resolving links. Default is the directory of the application entry point.
         :param uid: uid of the target instance
         """
-
-        if self._httpd:
-            self._httpd.shutdown()
 
         content = make_unicode(content)
         self.gui.load_html(content, base_uri, self.uid)
--- a/webview/http_server.py
+++ b/webview/http_server.py
@@ -1,25 +1,13 @@
-import webview
 import os
-import sys
+import logging
+from random import random
 import socket
 import threading
-import logging
+import wsgiref.simple_server
 
-from random import random
-
-try:
-    from BaseHTTPServer import HTTPServer
-    from SimpleHTTPServer import SimpleHTTPRequestHandler
-    from SocketServer import ThreadingMixIn
-except ImportError:
-    from http.server import SimpleHTTPRequestHandler, HTTPServer
-    from socketserver import ThreadingMixIn
-
-from webview.util import base_uri
 
 logger = logging.getLogger('pywebview')
 
-port = None
 
 def _get_random_port():
     def random_port():
@@ -28,62 +16,28 @@ def _get_random_port():
 
         try:
             sock.bind(('localhost', port))
-            result = port
-        except:
-            result = None
+        except OSError:
             logger.warning('Port %s is in use' % port)
+            return None
+        else:
+            return port
         finally:
             sock.close()
 
-        return result
-
-    port = random_port()
-    while not port:
-        port = random_port()
-
-    return port
-
-
-class HTTPHandler(SimpleHTTPRequestHandler):
-    def translate_path(self, path):
-        path = SimpleHTTPRequestHandler.translate_path(self, path)
-        relpath = os.path.relpath(path, os.getcwd())
-        fullpath = os.path.join(self.server.base_path, relpath)
-        return fullpath
-
-    def log_message(self, format, *args):
-        if os.environ.get('PYWEBVIEW_LOG') == 'debug':
-            super(HTTPHandler, self).log_message(format, *args)
-
-
-class ThreadedHTTPServer(ThreadingMixIn, HTTPServer):
-    """Handle requests in a separate thread."""
+    for port in iter(random_port, ...):
+        if port is not None:
+            return port
 
 
 def start_server(url):
-    def _start(httpd):
-        try:
-            httpd.serve_forever()
-        except Exception as e:
-            logger.exception(e)
-
-    global port
-
-    base_path = os.path.dirname(url.replace('file://', ''))
-    if not os.path.exists(base_path):
-        raise IOError('Directory %s is not found' % base_path)
-
     port = _get_random_port()
-    server_address = ('localhost', port)
+    server = wsgiref.simple_server.make_server('localhost', port, wsgiref.simple_server.demo_app)
 
-    httpd = ThreadedHTTPServer(server_address, HTTPHandler)
-    httpd.base_path = base_path
-
-    t = threading.Thread(target=_start, args=(httpd,))
+    t = threading.Thread(target=server.serve_forever)
     t.daemon = True
     t.start()
 
     new_url = 'http://localhost:{0}/{1}'.format(port, os.path.basename(url))
     logger.debug('HTTP server started on http://localhost:{0}'.format(port))
 
-    return new_url, httpd
+    return new_url
--- a/webview/platforms/winforms.py
+++ b/webview/platforms/winforms.py
@@ -284,7 +284,7 @@ class BrowserView:
             if self.httpd:
                 self.httpd.shutdown()
 
-            url, _ = start_server('file://' + self.temp_html)
+            url = start_server('file://' + self.temp_html)
             self.ishtml = True
             self.web_view.Navigate(url)
 


astronouth7303:
Note that this is a draft PR. I'm just trying to get it up earlier in case there's protests with the design.

astronouth7303:
TODO:

* [ ] winforms

astronouth7303:
And the following code works:

```python
import logging
import webview
from webview.http_server import Routing, StaticFiles

logging.basicConfig(level='DEBUG')

app = Routing({
    '/': StaticFiles('examples/assets'),
    '/code': StaticFiles('src'),
})

webview.create_window('test', app)
webview.start()
```

astronouth7303:
I'm also not super-happy with the names, so suggestions would be appreciated.

astronouth7303:
Hm. There's something funky going on with the range requests. It generally works, but occasionally it just fails?

astronouth7303:
correction: there's something funky going on with the wsgiref server.

astronouth7303:
Ok, I think this is ready for initial review. (I really don't expect it to pass.)

The only problem is that when using `StaticFiles` with `<audio>` on gtk, it sometimes just never loads? I'm not able to determine what's blocking--it doesn't _seem_ like the server, but I just can't tell.

astronouth7303:
I also seem to be able to get it to happen most reliably if i try to advance through a playlist quickily.

appveyor.yml
```diff
@@ -57,7 +57,7 @@ install:
   # compiled extensions and are not provided as pre-built wheel packages,
   # pip will build them from source using the MSVC compiler matching the
   # target Python version and architecture
-  - "%CMD_IN_ENV% pip install \"pytest>=4.3.0\" pythonnet six cefpython3==66.0"
+  - "%CMD_IN_ENV% pip install \"pytest>=4.3.0\" .[cef]"
```
r0x0r, line 60: What does this syntax `.[cef]`do?

examples/flask_app/src/backend/main.py
```diff
@@ -3,44 +3,14 @@
...
-        t.daemon = True
-        t.start()
-        logger.debug('Checking server')
-
-        while not url_ok('127.0.0.1', 23948):
-            sleep(1)
-
-        logger.debug('Server started')
-        window = webview.create_window('My first pywebview application', 'http://127.0.0.1:23948')
+        window = webview.create_window('My first pywebview application', server)
```
r0x0r, line 15: This is much more clear than before. Good work.

webview/serving.py
```diff
@@ -0,0 +1,134 @@
...
+                sock.bind(('localhost', port))
+            except OSError:
+                logger.warning('Port %s is in use' % port)
+                continue
+            else:
+                return port
+
+
+class WSGIRequestHandler11(wsgiref.simple_server.WSGIRequestHandler):
+    protocol_version = "HTTP/1.1"
```
r0x0r, line 37: Could you change this to single quotes?

webview/window.py
```diff
@@ -49,7 +49,8 @@ def __init__(self, uid, title, url, html, width, height, x, y, resizable, fullsc
                  background_color, js_api, text_select):
         self.uid = uid
         self.title = make_unicode(title)
-        self.url = None if html else transform_url(url)
+        self.url = None if html else url
+        self.real_url = None
```
r0x0r, line 53: What are the roles of `self.url` and `self.real_url`? Is the latter supposed to be public?
The `self.url` is a bit of a misnomer, as it does not necessarily reflect the current URL. It is not documented, but should be probably be private.

webview/wsgi.py
```diff
@@ -0,0 +1,423 @@
...
+        slashed, basename = posixpath.split(file)
+        slashed = slashed.rstrip('/')
+        if slashed:
+            packagename = "{}.{}".format(self.root, slashed.replace('/', '.'))
+        else:
+            packagename = self.root
+        try:
+            return importlib_resources.open_binary(packagename, basename)
+        except ModuleNotFoundError:
+            raise FileNotFoundError
```
r0x0r, line 423: `wsgi.py` and `serving.py` could be moved to the `http` subdirectory.

r0x0r:
@astronouth7303 Considering this is a big chance, would be able to stick around at least for some time to address future bugs?

astronouth7303:
Yeah, I'm extremely active on GitHub, so just @ me if you need help.

The streaming stall is a concern--i hit it occasionally during normal playback. But I'm not entirely sure what's going on.

r0x0r:
`examples/http_server.py` does not work on Cocoa. Displays a blank screen.

webview/window.py
astronouth7303: `url` is the thing the user passed in, which might be a URL (`str`), a local path (`str` or `Path`), or a WSGI app (callable).

`real_url` is an actual URL (`str`) that browsers will understand.

appveyor.yml
astronouth7303: Installs the local directory as a package with the extra `cef`.

astronouth7303: aka, DRYs your dependency lists.

webview/wsgi.py
astronouth7303: Yeah, thought about that, not sure it would help anything? The names seem descriptive and there's not really any clutter.

astronouth7303:
I only use linux and don't own a mac at all, so I can't really help you test that?

It might be because I haven't updated the cocoa and winforms backends for the `url`/`real_url` distinction.

webview/window.py
r0x0r: `window.url` is not used anywhere apart from the initialisation, nor documented, so instead of creating a `window.real_url` how about reusing `window.url` and renaming it to `window._url`?

webview/wsgi.py
r0x0r: It is a personal preference, but I am fine either way. If this implementation grows beyond two files, then it would warrant an own directory.

r0x0r:
Ok, I will test on other platforms. Have you tested on QT/GTK?

astronouth7303:
Yeah, GTK primarily but works in QT.

I can do a blind update for Cocoa/winforms, too. (It's mostly replacing `window.url` with `window.real_url`.)

r0x0r:
I get bunch of exceptions with `examples/http_server.py` on QT. The page renders though. Any ideas?

`
127.0.0.1 - - [02/Jun/2020 23:48:16] "GET /video.mp4 HTTP/1.1" 206 331776
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 138, in run
    self.finish_response()
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 180, in finish_response
    self.write(data)
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 279, in write
    self._write(data)
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 453, in _write
    result = self.stdout.write(data)
  File "/usr/lib/python3.6/socketserver.py", line 803, in write
    self._sock.sendall(b)
BrokenPipeError: [Errno 32] Broken pipe
127.0.0.1 - - [02/Jun/2020 23:48:16] "GET /video.mp4 HTTP/1.1" 500 59
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 43734)
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 138, in run
    self.finish_response()
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 180, in finish_response
    self.write(data)
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 279, in write
    self._write(data)
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 453, in _write
    result = self.stdout.write(data)
  File "/usr/lib/python3.6/socketserver.py", line 803, in write
    self._sock.sendall(b)
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.6/wsgiref/handlers.py", line 141, in run
    self.handle_error()
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 368, in handle_error
    self.finish_response()
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 180, in finish_response
    self.write(data)
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 274, in write
    self.send_headers()
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 331, in send_headers
    if not self.origin_server or self.client_is_modern():
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 344, in client_is_modern
    return self.environ['SERVER_PROTOCOL'].upper() != 'HTTP/0.9'
TypeError: 'NoneType' object is not subscriptable

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.6/socketserver.py", line 320, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/lib/python3.6/socketserver.py", line 351, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python3.6/socketserver.py", line 364, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.6/socketserver.py", line 724, in __init__
    self.handle()
  File "/usr/lib/python3.6/wsgiref/simple_server.py", line 133, in handle
    handler.run(self.server.get_app())
  File "/usr/lib/python3.6/wsgiref/handlers.py", line 144, in run
    self.close()
  File "/usr/lib/python3.6/wsgiref/simple_server.py", line 35, in close
    self.status.split(' ',1)[0], self.bytes_sent
AttributeError: 'NoneType' object has no attribute 'split'
`

astronouth7303:
I think `BrokenPipeError` means "the client closed the connection". I think we can safely shush that. (Although actually doing so sounds like it's going to involve reaching into places I was hoping to avoid reaching in to.)

Not sure what the `TypeError` is about. `environ` is set by the initial wsgi call, so it should never be `None`.

The `AttributeError` I would guess comes from if the connection is closed before the response is sent?

webview/window.py
astronouth7303: Ok, I can do that.

r0x0r:
It would be best to hide these errors. If you can find a way, that would be great. If it is reported via a logger, it can be hidden by overriding the logger. 
On that note, it would be nice to provide a way to hide requests as well. Maybe hide requests by default and show them only when `PYWEBVIEW_LOG=debug` is set. That could be a matter for another PR.

webview/window.py
r0x0r: On second thoughts, let's leave it as is. I have renamed window.url to window.original_url.

r0x0r:
@astronouth7303 Please give me write permissions to update your branch.

astronouth7303:
"Allow edits by maintainers" is checked.

Sorry, I've got like 4 things going on.

r0x0r:
@astronouth7303 Are you going to make any changes to this in the near future? If not, I will merge it.

PR #520, opened 2020-05-18
Frameless drag region

Fizzadar:
Finally figured this out! Allows user-defined "window drag regions" without per-platform hacks (like the cocoa `DragBar`).

If all good will update the other platforms.


r0x0r:
Works beautifully on Cocoa. Some words on the architecture and API. I thin the best way forward would be to incorporate dragging logic into pywebview code and introduce a magic class that would trigger drag logic (e.g. `pywebview-drag-region` or something like this). Class name could be made configurable too (`webview.DRAG_REGION_CLASS_NAME` for example).

I think it would be beneficial to keep the current logic of full window area drag in place, but make it optional (`create_window(framless=True, easy_drag=False)`). `easy_drag` would be False by default, which would break backward compatibility, but in my opinion this is a minor matter. 

Thoughts?

Fizzadar:
I think this sounds great - only thing is I think the default for `easy_drag` should remain `True` (at least for v3.x compatibility) and change it in v4.

For the JS itself I'll follow the same logic as the other JS bits, and trigger the event on any element matching the class name. How should the `moveWindow` function be added? Seems like something that would be good to keep separate to the JS API to avoid clashing method names.

r0x0r:
Yeah, good point about backward compatibility. 

The function name could be `window.pywebview._moveWindow`. It does not have to be under `window.pywebview.api`

Fizzadar:
I have implemented the `easy_drag` kwarg and other changes, unsure about https://github.com/r0x0r/pywebview/pull/520/commits/15c3a0725cd2e33640b76e8fc815309a3989eb8a but otherwise should be good to go?

webview/util.py
```diff
@@ -131,6 +136,10 @@ def _call():
 
         window.evaluate_js(code)
 
+    if window.frameless and not window.easy_drag and func_name == 'moveWindow':
```
r0x0r, line 139: Imo drag region can be enabled by default as long as the region class is found. Frameless window should not be a prerequisite to it.
Otherwise looks good. Great job!

Fizzadar: Agreed, have enabled it by default in https://github.com/r0x0r/pywebview/pull/520/commits/1d7a5159cc6460aaaa8da5f5ea80ef31c206b104.

r0x0r:
I get this on GTK. 

`Traceback (most recent call last):
  File "/home/roman/Code/pywebview/webview/platforms/gtk.py", line 215, in on_title_change
    value_id = js_data['id']
KeyError: 'id'`

r0x0r:
QT returns 
`
  File "/home/roman/Code/pywebview/webview/platforms/qt.py", line 84, in call
    return js_bridge_call(self.window, func_name, param, value_id)
  File "/home/roman/Code/pywebview/webview/util.py", line 137, in js_bridge_call
    window.move(*param)
  File "/home/roman/Code/pywebview/webview/window.py", line 31, in wrapper
    return function(*args, **kwargs)
TypeError: move() missing 2 required positional arguments: 'x' and 'y'
`

r0x0r:
Does not work with MSHTML too.

r0x0r:
forEach is the culprit for MSHTML

Fizzadar:
https://github.com/r0x0r/pywebview/pull/520/commits/0a7d4590eddf5a6121664ec8d4775ea56c2c6e95 should fix the GTK issue (encountered this myself on my own project).

And https://github.com/r0x0r/pywebview/pull/520/commits/fb3fd1625cbf8e8181768748bbbba72fddf90945 should cover the MSHTML issue.

Currently working on getting a test environment setup for QT to fix that.

Would also like to know your thoughts on exposing the function to make the binds, perhaps `pywebview.addDragRegions` or similar; I'm working on a React based app and have currently copied over the function into my codebase, but would be nice to reference the JS here.

r0x0r:
@Fizzadar I have fixed QT. Could you give me push permissions?

Fizzadar:
@r0x0r I've added you as a collaborator, thanks for fixing QT!

PR #598, opened 2020-09-03
Break Travis Linux tests in favor of the truth.

ryneeverett:
This breaks the build because the tests were already failing, the failure
was just being suppressed. Travis would only fail if the last pytest
invocation in run.sh had a non-zero exit code. It appears this has been
the case since 3.0.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -6,55 +6,65 @@ PYTEST_OPTIONS='-s --disable-pytest-warnings -r w'
 cd "${0%/*}"
 rm -r __pycache__ || true
 
-pytest test_bg_color.py::test_bg_color ${PYTEST_OPTIONS}
-pytest test_bg_color.py::test_invalid_bg_color ${PYTEST_OPTIONS}
+exitcode=0
+pywebviewtest() {
+  pytest "$@" || exitcode=$?
+}
 
-pytest test_evaluate_js.py::test_mixed ${PYTEST_OPTIONS}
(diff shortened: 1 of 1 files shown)


r0x0r:
Truth is welcomed. Thanks for this fix.

I get this error on Ubuntu 18 with default bash.
./run.sh: line 9: var: command not found

ryneeverett:
Looks like I mixed up languages there. I force pushed a fix.

r0x0r:
@ryneeverett  The plot thickens. JS related tests throw this error when run with `run.sh`

```
  File "/home/roman/Code/pywebview/webview/platforms/gtk.py", line 192, in on_webview_ready
    self.shown.set()
  File "/home/roman/Code/pywebview/webview/event.py", line 33, in set
    if len(self._items):
```

But pass when run directly in a command line, eg.  `PYTHONPATH=.. PYWEBVIEW_LOG='debug' pytest test_evaluate_js.py::test_mixed -s --disable-pytest-warnings -r w`

paulsbond:
I get the same error:
`AttributeError: 'Event' object has no attribute '_items'`

With python 3.8.5 and pywebview-3.4

The following causes the error after closing the window:
```python
import webview

class RandomClass:
    def test(self, x):
        print(x)

if __name__ == "__main__":
    webview.create_window("Hello", "index.html")
    webview.start()
```
If I comment out the class or change it into a function the error goes away.
Nothing special in my index.html:

```html
<title>Test</title>
<h1>Test</h1>
```

PR #605, opened 2020-09-08
Fixes arrow keys not responding in text input fields.

awesomo4000:
Quick fix to make up down left and right work on OSX.  To see if you have this problem, try moving left and right in the text box of the todos example.


awesomo4000:
thanks! (Not sure if this is the best way to do this, but it worked for me)

PR #607, opened 2020-09-20
Fix AttributeError when called with html arg

dmwyatt:
`self.original_url` is None when `create_window` is called with the `html` argument.

Because of that this check would raise an AttributeError when it tried to call 
`startswith` on `self.original_url`.

### DIFF
--- a/webview/window.py
+++ b/webview/window.py
@@ -87,7 +87,14 @@ class Window:
         self._is_http_server = http_server
 
         # WebViewControl as of 5.1.1 crashes on file:// urls. Stupid workaround to make it work
-        if gui.renderer == "edgehtml" and (self.original_url.startswith('file://') or '://' not in self.original_url):
+        if (
+            gui.renderer == "edgehtml"
+            and self.original_url is not None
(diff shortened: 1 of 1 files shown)


dmwyatt:
The problem with this is that it doesn't actually seem to make edgehtml actually work with HTML.  

Before the fix in this PR, the code would just crash with the AttributeError.  After the fix in this PR, UAC prompts and the webview opens with the window title "Woah dude", but there is no content in the window, it's just white.

```python
import webview

webview.create_window('Woah dude!', html='<h1>Woah dude!<h1>')
webview.start()
```

r0x0r:
@dmwyatt 3.3.4 introduced a fix for loading file:// urls in EdgeHTML, but this broken loading html (as it uses a http server for loading html). Unfortunately your PR won't help, as the problem lies elsewhere. I will try to look into this.

r0x0r:
Pushed a fix to master. It should work both with loading html, as well as local URLs. 

@dmwyatt can you test it out?

PR #623, opened 2020-10-20
Auto Title

Ksengine:
Automatically set title of webpage as window title


r0x0r:
Please do not create feature pull requests without discussing the matter first. 

I have doubts about this feature. My main concerns are
1) Poor UX as window title is not available until page is loaded
2) Whether it is needed at all

Issue #627, opened 2020-10-28
Stop freezing Main thread


Ksengine:
can we stop freezing main thread.
Its working with gtk and qt.
not working on windows MSHTML
Cocoa not tested.

We can do it with multiprocessing.

r0x0r:
Are you suggesting to run GUI loop as a separate process and keep the main thread of the main process free?

Ksengine:
yes,
you added multiproessing as an argument of start method
but not supported yet

Ksengine:
I can help

r0x0r:
The original multiprocessing idea was to use internally processes instead
of threads. But I like your approach more. Feel free to go ahead with
implementation. The parameter name could be `webview.start(block=True)`.
The default value is True by default to keep it backward compatibility.

Ksengine:
ok i will add this feature

Ksengine:
can you add labels to this issue

r0x0r:
One consideration that should be decided is the exit logic. Normally the
program exits after the gui loop ends but with a non-blocking gui what
would be the default logic? With a non-blocking user code what is the
behaviour after the code finishes executing?

r0x0r:
I like the join syntax, as it is inline with the Python API. Let's name the parameter `daemon` instead of `block` and make `start` return a process object in the daemon mode. ie.

```
process = webview.start(daemon=True)
do_stuff()
process.join()
# app is exited after this
```

We need to think of a way to signal the gui process to quit after the join, as it would not happen automatically.

Ksengine:
i added my pull request.
**sorry, i hadn't read your last reply**
so i added both `daemon` and `block` methods
but they are different
i copied from #632 
> - multiprocessing added
> - fixed threading
> - added two keyword arguments to start method
>     - block(default : True)
>     - daemon(default : True)(works if block is False)
> 
> ex(block):-
> ```python
> import webview
> w = webview.create_window("ssss","https://www.google.com")
> join = webview.start(block=False) # set block option to False( default:True )
> print('free main thread')
> # Some code
> print('blocked main thread')
> join() # join gui loop to main thread
> print('free main thread') # this line works after gui loop ended
> ```
> 
> ex(block without join):-
> ```python
> import webview
> w = webview.create_window("ssss","https://www.google.com")
> join = webview.start(block=False) # set block option to False( default:True )
> print('free main thread')
> # Some code
> # program exits when no any code run and gui loop ended
> ```
> 
> ex(daemon):-
[...]

PR #632, opened 2020-10-31
Stop freezing Main thread

Ksengine:
- multiprocessing added
- fixed threading
- added two keyword arguments to start method
    - block(default : True)
    - daemon(default : True)(works if block is False)

ex(block):-
```python
import webview
w = webview.create_window("ssss","https://www.google.com")
join = webview.start(block=False) # set block option to False( default:True )
print('free main thread')
# Some code
print('blocked main thread')
join() # join gui loop to main thread
print('free main thread') # this line works after gui loop ended
```

ex(block without join):-
```python
import webview
w = webview.create_window("ssss","https://www.google.com")
join = webview.start(block=False) # set block option to False( default:True )
print('free main thread')
# Some code
# program exits when no any code run and gui loop ended
```

ex(daemon):-
```python
import webview
w = webview.create_window("ssss","https://www.google.com")
join = webview.start(block=False, daemon=False) # set daemon option to False( default:True )
print('free main thread')
# Some code
# gui loop stops and exit program
```


Ksengine:
according to this issue https://github.com/r0x0r/pywebview/issues/627

webview/__init__.py
```diff
@@ -96,24 +91,39 @@ def _create_children(other_windows):
...
-            t = Thread(target=func)
+            t = threading.Thread(target=func)
         t.start()
 
     guilib.create_window(windows[0])
 
 
+def start(*args,**kwargs):
+    block = kwargs.get('block', True)
+    daemon = kwargs.pop('daemon', True)
```
r0x0r, line 114: Let's have only one argument `daemon` that would be the new name for block. Ie. the default would be `daemon=False` (start is blocking) and with True it would be not block until joined.

webview/__init__.py
```diff
@@ -96,24 +91,39 @@ def _create_children(other_windows):
...
+    
+    if block:
+        real_start(*args, **kwargs)
+        
+    else:
+        mp.get_context('fork')
+        p=mp.Process(target=real_start, args=args, kwargs=kwargs)
```
r0x0r, line 121: Spaces around assign sign `p = mp.Process`

webview/__init__.py
```diff
@@ -96,24 +91,39 @@ def _create_children(other_windows):
...
+        p.daemon = daemon
```
r0x0r, line 122: daemon would be always True. It should be verified whether multiprocessing module is able to shutdown spawned GUI process automatically or whether additional quite logic is needed.

webview/__init__.py
```diff
@@ -64,7 +65,7 @@
 token = _token
 windows = []
 
-def start(func=None, args=None, localization={}, gui=None, debug=False, http_server=False, user_agent=None):
+def real_start(func=None, args=None, localization={}, gui=None, debug=False, http_server=False, user_agent=None, block=True):
```
r0x0r, line 68: `_start` is a more suitable name

webview/__init__.py
```diff
@@ -76,15 +77,9 @@ def _create_children(other_windows):
 
     _debug = debug
     _user_agent = user_agent
-    #_multiprocessing = multiprocessing
-    multiprocessing = False # TODO
+    _multiprocessing = block
```
r0x0r, line 80: This variable can be deleted.

webview/__init__.py
```diff
@@ -96,24 +91,39 @@ def _create_children(other_windows):
     guilib = initialize(gui)
 
     for window in windows:
-        window._initialize(guilib, multiprocessing, http_server)
+        window._initialize(guilib, _multiprocessing, http_server)
```
r0x0r, line 94: same here

r0x0r (changes requested):

webview/__init__.py
Ksengine: ok

webview/__init__.py
Ksengine: multiprocessing is a global variable
it is used by window objects

webview/__init__.py
Ksengine: window object requires this argument

webview/__init__.py
Ksengine: > daemon would be always True.
if daemon is True python kills gui loop automatically
**to avoid this users shod use something like this**
```python
 ...
webview.start(daemon = True)
print('free main thread')
while True:
    pass # keep gui loop
```
so i added both `block` and `daemon` args

webview/__init__.py
Ksengine: ok

webview/__init__.py
Ksengine: https://github.com/r0x0r/pywebview/pull/632#discussion_r515735465

webview/__init__.py
```diff
@@ -64,7 +65,7 @@
 token = _token
 windows = []
 
-def start(func=None, args=None, localization={}, gui=None, debug=False, http_server=False, user_agent=None):
+def _start(func=None, args=None, localization={}, gui=None, debug=False, http_server=False, user_agent=None, block=True):
```
Ksengine, line 68: resolved

Ksengine (commented):
renamed `real_start` to `_start`

webview/__init__.py
r0x0r: It is not used in any meaningful way in window.py, so it can be purged from there as well

webview/__init__.py
r0x0r: Sorry, daemon would be False by default, so it would block the main thread.
Join or other way (http server loop or something like that) would be require to prevent program termination. I see no value in letting implicitly child process be alive. 

By the way WkWebView on macOS uses multiprocessing architecture by default. It must be verified whether it works with daemonized multiprocessing. If not, then a custom solution to terminate gui loop is needed.

Ksengine:
sorry windows test failed

Ksengine:
sorry its a wrong method
dont merge this

Issue #521, opened 2020-05-19
Edge WebView2 support


r0x0r:
Windows Forms support for Chromium based WebView2 has finally arrived.  Implementing this for pywebview should be doable at this point.

https://docs.microsoft.com/en-us/microsoft-edge/webview2/gettingstarted/winforms

kvnxiao:
Any chance of this being done any time soon?

r0x0r:
@kvnxiao I am not working on this actively at the moment, not sure when I have time and energy to devote to this issue. 

However, if you want to see this implemented fast, consider sponsoring this issue.

kvnxiao:
Suppose if I have time to contribute to this, what exactly is required to get WebView2 up and running together?

r0x0r:
1) WebView2 implementation is to be done in the same manner as EdgeHTML, ie no changes would be needed to made to window logic. WebView2 component would be attached as a child to the form.

2) WebView2 detection logic is needed (checking for .NET and Window 10 versions ), as it is done with Edge https://github.com/r0x0r/pywebview/blob/746fd6a16fb678d4d63b4fe8cd61476154f5516d/webview/platforms/winforms.py#L68

3) `winforms.py` will require a refactor at some point, as it is already too large for a single file. In practice that would be mean splitting renderer code to separate files.

4) `renderer` value for WebView2 can be `chromium`

5) EdgeHTML is buggy af, not sure if WebView2 is any better. In any case, one should prepare for all kinds of quirks and funky solutions.

kvnxiao:
WebView2 uses Chromium so it should not be buggy in any way; it would be as buggy as using chromium on other platforms.

dmwyatt:
@kvnxiao Are you going to work on this?

kvnxiao:
@dmwyatt I don't think I have the time to contribute on this one, but just simply wanted to ask for a clear roadmap in case someone else would like to work on it.

tshemeng:
[Announcing Microsoft Edge WebView2 General Availability](https://blogs.windows.com/msedgedev/2020/10/19/edge-webview2-general-availability/)

Ksengine:
https://github.com/MicrosoftEdge/WebView2Samples/blob/master/SampleApps/WebView2WindowsFormsBrowser

**?**

Ksengine:
[webview2 latest](https://github.com/r0x0r/pywebview/files/5493612/zip.zip)
dlls

Ksengine:
[...]

PR #649, opened 2020-11-17
Edge WebView2 support 

sbbosco:
Implement Edge WebView2 support #521 
Adds Windows Forms support for Chromium based WebView2.
Refactor render classes to separate files.


r0x0r:
I tested this briefly and I get a blank screen. No errors.

Also none of those registry keys were found in registry, but instead there was a {56EB18F8-B008-4CBD-B6D2-8C97FE7E9062} with the pv value of 86.0.622.69. Tested with this value.

sbbosco:
Unfortunately, Microsoft has decided that WebView2 will not work with the stable version of Edge {56EB18F8-B008-4CBD-B6D2-8C97FE7E9062}.  It only works with one of the following: 
* The **WebView2 Runtime**. https://developer.microsoft.com/en-us/microsoft-edge/webview2/
* Any Insider (non-stable) Microsoft Edge (Chromium) browser channel such as Beta, Dev or Canary.

WebView2 distribution: https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/distribution

When I forced the **is_chromium** function to False, I defaulted to IE (mshtml). We may need change the logic to make EdgeHTML the fallback.
`is_edge = _is_edge() and forced_gui_ != 'mshtml' and not is_chromium`

What version of Windows 10 are you running? I am on 20H2.

Ksengine:
## to force install `WebView2 Runtime`
goto https://developer.microsoft.com/en-us/microsoft-edge/webview2/#download-section
then click `Get The Link`, accept license and copy your link
using link we can programmatically download WebView2 EverGreen Installer(less than 2 mb). just run it

r0x0r:
@sbbosco Thanks for the clarification. Will download the Edge runtime and test it again. I don't have access to a Windows development machine at the moment, using some IE W10 image inside a virtual machine instead.

I refactored the renderer loading logic so it will always prefer Chromium over Edge, and Edge over IE.

sbbosco:
@r0x0r Nice clean-up of the chromium registry check.

I modified the registry check for Windows 32-bit OS. I also fixed a "local variable 'windows_key' referenced before assignment" exception. I am running Edge Beta on my main machine instead of the runtime and the function was returning False.

webview/platforms/edgechromium.py
```diff
@@ -0,0 +1,204 @@
...
+            self.syncContextTaskScheduler)
+        except Exception as e:
+            logger.exception('Error occurred in script')
+            self.js_result = None
+            self.js_result_semaphore.release()
+
+    def get_current_url(self):
+        return self.url
+
+    def load_html(self, html, base_uri):
```
r0x0r, line 107: HTML is used for loading HTML in EdgeHTML, because of the bug in EdgeHTML which crashes regular html loading. I guess this is not a case in Chromium, so regular html loading can be used instead.

webview/platforms/edgechromium.py
```diff
@@ -0,0 +1,204 @@
...
+        #settings under on_webview_ready 
+        self.web_view.CoreWebView2Ready += self.on_webview_ready
+        self.web_view.NavigationStarting += self.on_navigation_start
+        self.web_view.NavigationCompleted += self.on_navigation_completed
+        self.web_view.WebMessageReceived += self.on_script_notify
+
+        self.httpd = None # HTTP server for load_html
+        self.tmpdir = None
+        self.url = None
+        self.ishtml = False
```
r0x0r, line 69: All the http server related stuff can be purged

webview/platforms/edgechromium.py
```diff
@@ -0,0 +1,204 @@
...
+            logger.exception('Exception occured during on_script_notify')
+
+    def on_new_window_request(self, _, args):
+        args.set_Handled(True)
+        #webbrowser.open(str(args.get_Uri()))
+
+    def on_webview_ready(self, sender, args):
+        sender.CoreWebView2.NewWindowRequested += self.on_new_window_request
+        settings = sender.CoreWebView2.Settings
+        if not _debug: settings.AreDefaultContextMenusEnabled = False
```
r0x0r, line 145: settings.AreDefaultContextMenusEnabled = _debug

webview/platforms/edgechromium.py
```diff
@@ -0,0 +1,204 @@
...
+    def on_navigation_start(self, sender, args):
+        pass
+
+    def on_navigation_completed(self, sender, args):
+        try:
+            if self.tmpdir and os.path.exists(self.tmpdir):
+                shutil.rmtree(self.tmpdir)
+                self.tmpdir = None
+        except Exception as e:
+            logger.exception('Failed deleting %s' % self.tmpdir)
```
r0x0r, line 163: http server stuff

webview/platforms/edgechromium.py
Ksengine: yes

webview/platforms/edgechromium.py
sbbosco: The http server related stuff has been purged. Settings tweaked for _debug.

webview/platforms/edgechromium.py
sbbosco: The http server related stuff has been purged,

webview/platforms/edgechromium.py
sbbosco: Settings tweaked for _debug.

r0x0r:
Tests failing, at least `test_load_html` and `test_multi_window`

```
    def test_load_html():
        window = webview.create_window('Load HTML test')
>       run_test(webview, window, load_html)
E       Failed: Traceback (most recent call last):
E         File "Z:\Code\pywebview\tests\util.py", line 65, in thread
E           thread_func(window)
E         File "Z:\Code\pywebview\tests\test_load_html.py", line 11, in load_html
E           window.load_html('<h1>This is dynamically loaded HTML</h1>')
E         File "Z:\Code\pywebview\webview\window.py", line 31, in wrapper
E           return function(*args, **kwargs)
E         File "Z:\Code\pywebview\webview\window.py", line 181, in load_html
E           self.gui.load_html(content, base_uri, self.uid)
E         File "Z:\Code\pywebview\webview\platforms\winforms.py", line 555, in load_html
E           BrowserView.instances[uid].load_html(content, base_uri)
E         File "Z:\Code\pywebview\webview\platforms\winforms.py", line 268, in load_html
E           self.Invoke(Func[Type](_load_html))
E         File "Z:\Code\pywebview\webview\platforms\winforms.py", line 266, in _load_html
E           self.browser.load_html(content, base_uri)
E         File "Z:\Code\pywebview\webview\platforms\edgechromium.py", line 107, in load_html       
E           sender.CoreWebView2.NavigateToString(html)
E       AttributeError: 'str' object has no attribute 'CoreWebView2'
```

Ksengine:
Sender is a str object

sbbosco:
I didn't realize **load_html** would be run directly. I had redefined the parameters. The **test_load_html** and **test_multi_window** tests should work now. 

When the WebView2 control is created, the CoreWebView2 property is null and loads as a asynchronous task. You can't set HTML until the asynchronous task has completed.

r0x0r:
Looks good now. Resize test fails but I guess this is due poor performance of a virtual machine.

I will update the documentation. Also appveyor config should be checked on whether it is possible to install Edge runtime there.

Ksengine:
> ## to force install `WebView2 Runtime`
> goto https://developer.microsoft.com/en-us/microsoft-edge/webview2/#download-section
> then click `Get The Link`, accept license and copy your link
> using link we can programmatically download WebView2 EverGreen Installer(less than 2 mb). just run it

:question:

r0x0r:
That's not a good idea and is better left to user/third party code.
A more general solution would be to introduce  a `webview.available_renderers()` function that would report which renderers are available before creating a window.

webview/platforms/winforms.py
```diff
@@ -17,15 +16,20 @@
...
 from ctypes import windll
 from uuid import uuid4
+from platform import machine
+import time
 
 from webview import WebViewException, windows, OPEN_DIALOG, FOLDER_DIALOG, SAVE_DIALOG, _debug, _user_agent
 from webview.guilib import forced_gui_
-from webview.serving import resolve_url
-from webview.util import parse_api_js, interop_dll_path, parse_file_type, inject_base_uri, default_html, js_bridge_call
+from webview.util import parse_file_type, inject_base_uri
```
Ksengine, line 24:
```suggestion
from webview.util import parse_file_type, inject_base_uri, Process
```
**NameError**

Ksengine (commented):
Errs

Ksengine:
err found
```python
  File "<stdin>", line 1, in <module>
  File "C:\Users\Kavindu\Downloads\pywebview-master\pywebview-master\webview\__i
nit__.py", line 90, in start
    guilib = initialize(gui)
  File "C:\Users\Kavindu\Downloads\pywebview-master\pywebview-master\webview\gui
lib.py", line 104, in initialize
    if not try_import(guis):
  File "C:\Users\Kavindu\Downloads\pywebview-master\pywebview-master\webview\gui
lib.py", line 68, in try_import
    if import_func():
  File "C:\Users\Kavindu\Downloads\pywebview-master\pywebview-master\webview\gui
lib.py", line 58, in import_winforms
    import webview.platforms.winforms as guilib
  File "C:\Users\Kavindu\Downloads\pywebview-master\pywebview-master\webview\pla
tforms\winforms.py", line 139, in <module>
    from . import mshtml as IE
  File "C:\Users\Kavindu\Downloads\pywebview-master\pywebview-master\webview\pla
tforms\mshtml.py", line 36, in <module>
    clr.AddReference(interop_dll_path('WebBrowserInterop.dll'))
System.IO.FileNotFoundException: Unable to find assembly 'C:\Users\Kavindu\Downl
oads\pywebview-master\pywebview-master\webview\lib\WebBrowserInterop.x86.dll'.
   at Python.Runtime.CLRModule.AddReference(String name)
```

sbbosco:
@Ksengine I'm not sure why you are trying to run this from Downloads instead of from a normal python install via ```pip install pywebview```, but the dll may blocked because it is an internet download. Right-Click **WebBrowserInterop.x86.dll** and check unblock.
https://github.com/sbbosco/pywebviewx/blob/main/webview/lib/unblock.png

Ksengine:
> @Ksengine I'm not sure why you are trying to run this from Downloads instead of from a normal python install via `pip install pywebview`, but the dll may blocked because it is an internet download. Right-Click **WebBrowserInterop.x86.dll** and check unblock.
> https://github.com/sbbosco/pywebviewx/blob/main/webview/lib/unblock.png

I am testing `noblock`

divadsn:
> That's not a good idea and is better left to user/third party code. A more general solution would be to introduce a `webview.available_renderers()` function that would report which renderers are available before creating a window.

Any plan or idea on that one? I would want to force fully exit the app when edgechromium is not available, right now pywebview is falling back to mshtml which is far from ideal to be honest

r0x0r:
> > That's not a good idea and is better left to user/third party code. A more general solution would be to introduce a `webview.available_renderers()` function that would report which renderers are available before creating a window.
> 
> Any plan or idea on that one? I would want to force fully exit the app when edgechromium is not available, right now pywebview is falling back to mshtml which is far from ideal to be honest

`webview.available_renderers()` is not something I am going to implement, but instead I have introduced `window.events.initialized` event. It sends a renderer is a param and by returning False in an event handler you can cancel initialization. You can find the implementation in the latest commit of the `6.0` branch

PR #638, opened 2020-11-04
Stop freezing main thread

Ksengine:
stop freezing main thread
- winforms Edge(supported, tested)
- winforms CEF (supported, not tested)
- winforms IE (supported, not tested)
- gtk(supported, tested)
- qt(supported, tested on linux only)
- cocoa(not supported)

so,
- test IE
- test CEF
- test qt on mac
- develop cocoa

other developers are welcomed

PR #650, opened 2020-11-18
Non-blocking example

r0x0r:
Update example and documentation with information on how pywebview can be used in a non-blocking mode.

### DIFF
--- a/docs/examples/no_block.md
+++ b/docs/examples/no_block.md
@@ -1,0 +1,32 @@
+# Non-blocking architecture
+
+
+Create a pywebview window in a separate process, so that main thread is not blocked.
+
+
+``` python
+import webview
+import multiprocessing
+import sys
+
+
+if getattr(sys, 'frozen', False):
+    multiprocessing.freeze_support()
+
+
+def create_window():
+    window = webview.create_window('Simple browser', 'https://pywebview.flowrl.com/hello')
+    webview.start()
+
+
+if __name__ == '__main__':
+    multiprocessing.set_start_method('spawn')
+    p = multiprocessing.Process(target=create_window)
+    p.start()
+
+    print('Main thread is not blocked. You can execute your code after pywebview window is created')
+
+    # main thread must be blocked using join or any other user defined method to prevent app from quitting
+    p.join()
+    print('App exists now')
+```
--- a/docs/guide/usage.md
+++ b/docs/guide/usage.md
@@ -67,6 +67,9 @@ webview.start(custom_logic, window)
 pass
 ```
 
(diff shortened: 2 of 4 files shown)


docs/examples/no_block.md
```diff
@@ -0,0 +1,32 @@
...
+    multiprocessing.freeze_support()
+
+
+def create_window():
+    window = webview.create_window('Simple browser', 'https://pywebview.flowrl.com/hello')
+    webview.start()
+
+
+if __name__ == '__main__':
+    multiprocessing.set_start_method('spawn')
```
Ksengine, line 23: remove this line.
on MacOS and Windows it will use 'spawn'
on Linux it will use `fork`(better method)
when frozen it will use `fork`

examples/no_block.py
```diff
@@ -0,0 +1,24 @@
...
+if __name__ == '__main__':
+    multiprocessing.set_start_method('spawn')
+    p = multiprocessing.Process(target=create_window)
+    p.start()
+
+    print('Main thread is not blocked. You can execute your code after pywebview window is created')
+
+    # main thread must be blocked using join or any other user defined method to prevent app from quitting
+    p.join()
+    print('App exists now')
```
Ksengine, line 24: No newline at end of file

Ksengine:
```suggestion
    print('App exists now. Main Thread is blocked.')
```

docs/examples/no_block.md
```diff
@@ -0,0 +1,32 @@
...
+    p = multiprocessing.Process(target=create_window)
+    p.start()
+
+    print('Main thread is not blocked. You can execute your code after pywebview window is created')
+
+    # main thread must be blocked using join or any other user defined method to prevent app from quitting
+    p.join()
+    print('App exists now')
```
Ksengine, line 31:
```suggestion
    print('App exists now. Main Thread is blocked.')
```

docs/guide/usage.md
```diff
@@ -67,6 +67,9 @@ webview.start(custom_logic, window)
 pass
 ```
 
+If you wish to keep main thread free, you can run pywebview in a separate process. Refer to [this example](/examples/no_block.md) for details. Note that in a non-blocking mode main thread must be blocked in one way or another. Otherwise the program will quit immediately after reaching the end of main function.
```
Ksengine, line 70: wrong.
> program will quit immediately after reaching the end of main function

process is not demon.
make it daemon to valid this line
```python
...Process(..., daemon=True)
```

but its a bad option because we can stop process using `Process.close()`

Ksengine (changes requested):
Bad Idea!!!

multiprocessing is an expensive method.
its only used on GTK
- qt - no GUI loop
- winforms - .NET Thread

Ksengine:
visit this [comment](https://github.com/r0x0r/pywebview/pull/638#issuecomment-730409921) in [#638](https://github.com/r0x0r/pywebview/pull/638#issuecomment-730409921)

PR #670, opened 2020-12-05
Example of usage with pystray

AlexCovizzi:


examples/pystray_icon.py
```diff
@@ -0,0 +1,58 @@
...
+
+"""
+This example demonstrates running pywebview alongside with pystray to display a system tray icon.
+"""
+
+
+def run_webview():
+    window = webview.create_window('Webview', 'https://pywebview.flowrl.com/hello')
+    webview.start()
+    return window
```
Ksengine, line 22: Returns window after it closed

r0x0r:
@AlexCovizzi I have tested it on macOS and Windows.

**macOS:** 
does not start, shows 
```
The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec().
Break on __THE_PROCESS_HAS_FORKED_AND_YOU_CANNOT_USE_THIS_COREFOUNDATION_FUNCTIONALITY___YOU_MUST_EXEC__() to debug.
```
Setting multiprocessing.set_start_method('spawn') remedies it a bit. At least pywebview start, but trayicon is nowhere to be seen.

**Windows**
Works as expected, but the Exit button in the tray menu only destroys the tray icon, not the window.

examples/pystray_icon.py
AlexCovizzi: fixed it

AlexCovizzi:
@r0x0r For Mac OS I don't really have a way to test it at the moment, I added `multiprocessing.set_start_method('spawn') ` as suggested
For the other issue I kept it like that on purpose otherwise the example would be a bit more complicated

Ksengine:
> @r0x0r For Mac OS I don't really have a way to test it at the moment, I added `multiprocessing.set_start_method('spawn') ` as suggested
> For the other issue I kept it like that on purpose otherwise the example would be a bit more complicated

MacOs is using fork as default method

r0x0r:
@AlexCovizzi It would be better to remove the macOS support for now and add a comment that macOS is not supported or something along these lines. If somebody ever needs it, they can figure it out themselves.

examples/pystray_icon.py
```diff
@@ -0,0 +1,56 @@
...
+    icon.run()
+
+
+if __name__ == '__main__':
+    queue = Queue()
+ 
+    icon_thread = Thread(target=run_pystray, args=(queue,))
+    icon_thread.start()
+
+    run_webview()
```
r0x0r, line 44: The gui thread blocks at this point. The code below will execute after pywebview window is closed

examples/pystray_icon.py
```diff
@@ -1,16 +1,13 @@
...
+
 if sys.platform == 'darwin':
-    # System tray icon needs to run in it's own process on Mac OS X
-    import multiprocessing
-    from multiprocessing import Process as Thread, Queue
-    multiprocessing.set_start_method('spawn')
-else:
-    from threading import Thread
-    from queue import Queue
+    raise NotImplementedError('This example does not work on macOS.')
```
Ksengine, line 7:
```suggestion
    from multiprocessing import get_context
    ctx = get_context('spawn')
    Process = ctx.Process
    Queue = ctx.Queue
```

Ksengine (commented):
suggestion for MacOs

Ksengine:
# here is a real example
```python
"""
This example demonstrates running pywebview alongside with pystray to display a system tray icon.
"""
from PIL import Image
from pystray import Icon, Menu, MenuItem
import webview
import sys

if sys.platform == 'darwin':
    from multiprocessing import get_context
    ctx = get_context('spawn')
    Process = ctx.Process
    Queue = ctx.Queue

from threading import Thread
from queue import Queue

queue = Queue()
window = webview.create_window('Webview', 'https://pywebview.flowrl.com/hello')
hide = False


def event_loop():
    global hide, window
    while True:
        event = queue.get()
        if event == 'open':
            if hide:
                window.show()
                hide = False

            else:
                window.hide()
                hide = True

        if event == 'exit':
            break


def run_pystray(queue: Queue):
    def on_open(icon, item):
        queue.put('open')

    def on_exit(icon, item):
        icon.stop()
        queue.put('exit')

    image = Image.open('logo/logo.png')
    menu = Menu(MenuItem('Open/Hide', on_open), MenuItem('Exit', on_exit))
    icon = Icon('Pystray', image, "Pystray", menu)
    icon.run()


if __name__ == '__main__':
    icon_thread = Thread(target=run_pystray, args=(queue, ))
    tread.daemon = True
    icon_thread.start()

    webview.start(event_loop)

    icon_thread.join()
```

AlexCovizzi:
@Ksengine Do both pywebview and pystray work on Mac OS if you use `ctx = get_context('spawn')`?

There is one problem with that solution though, if you close the window and then click open it will throw an error, I think the proper solution would be to have 2 event loops (and 2 queues), one in a thread beside pywebview to handle the window destroy and one in the main thread that handles the the window creation, but I feel the example is going to become way to complicated that way

Ksengine:
> @Ksengine Do both pywebview and pystray work on Mac OS if you use ctx = get_context('spawn')?

see python docs

Ksengine:
> There is one problem with the code you posted, if you close the window and then click open it will throw an error, I think the proper solution would be to have 2 event loops (and 2 queues), one in a thread beside pywebview to handle the window destroy and one in the main thread that handles the the window creation, but I feel the example is going to become way to complicated that way

I had edited example

Ksengine:
@AlexCovizzi I think now you can edit your pr

AlexCovizzi:
I updated the example, I took a different approach: start pystray in the main process and spawn a new process for webview. Works very well on Windows, not sure about Mac OS though

@Ksengine The updated example that you posted was still not behaving as expected since once closed the webview it could not be reopend with the system tray icon

Ksengine:
:superhero::+1::+1::+1:

r0x0r:
@AlexCovizzi The updated example works perfectly on macOS. Thanks for the contribution.

Ksengine:
# @AlexCovizzi Its :cool:

PR #679, opened 2020-12-25
js bridge serialization fixes and optimization

peter23:
Hello.

1) In `js/drag.py` there is a call `window.pywebview._bridge.call('moveWindow', [x, y], 'move');`.
Then, if an engine is 'mshtml'/'cef'/'qtwebkit' then `window.external.call(funcName, params, id);` is called in `js/api.py`.
With 'mshtml' that causes an error: `param` is an object of type 'System.__ComObject' and call `window.move(*param)` in `util.py` fails with exception:
```
COMException : Error HRESULT E_FAIL has been returned from a call to a COM component.
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
   at System.Runtime.InteropServices.CustomMarshalers.EnumeratorViewOfEnumVariant.MoveNext()
   at System.Collections.IEnumerator.MoveNext()
   at Python.Runtime.Iterator.tp_iternext(IntPtr ob)
   at window.pywebview._bridge.call (eval code:420:21)
   at onMouseMove (eval code:1299:9)
```

2) About the same call in `js/api.py` (`window.external.call...`).
It is used for 'qtwebkit', but another call is used for 'qtwebengine', with `JSON.stringify(params)`.
But `platforms/qt.py` waits for json-encoded `param` in any case.

Solution for problems 1 and 2 is just to use json-encoding of `params` in any case.
(And `platforms/cef.py` and `platforms/mshtml.py` should be ready for that, another platforms already use json-encoding)
(It is done in `js/api.py` line 31, `platforms/mshtml.py` line 50, `platforms/cef.py` line 73)

3) Now, if we do json-encoding in `window.pywebview._bridge.call` in any case then we do not need additional json-encoding in `window.pywebview.api`.
(It is done in `js/api.py` line 17, `util.py` line 160)

### DIFF
--- a/webview/js/api.py
+++ b/webview/js/api.py
@@ -14,7 +14,7 @@ window.pywebview = {
                 "var promise = new Promise(function(resolve, reject) { " +
                     "window.pywebview._checkValue('" + funcName + "', resolve, reject, __id); " +
                 "}); " +
-                "window.pywebview._bridge.call('" + funcName + "', JSON.stringify(arguments), __id); " +
+                "window.pywebview._bridge.call('" + funcName + "', arguments, __id); " +
                 "return promise;"
 
             window.pywebview.api[funcName] = new Function(params, funcBody)
@@ -28,7 +28,7 @@ window.pywebview = {
                 case 'mshtml':
                 case 'cef':
                 case 'qtwebkit':
-                    return window.external.call(funcName, params, id);
+                    return window.external.call(funcName, JSON.stringify(params), id);
                 case 'edgehtml':
                     return window.external.notify(JSON.stringify([funcName, params, id]));
                 case 'chromium':
--- a/webview/platforms/cef.py
+++ b/webview/platforms/cef.py
@@ -70,7 +70,7 @@ class JSBridge:
         self.eval_events[uid].set()
 
     def call(self, func_name, param, value_id):
-        js_bridge_call(self.window, func_name, param, value_id)
+        js_bridge_call(self.window, func_name, json.loads(param), value_id)
 
 
 renderer = 'cef'
--- a/webview/platforms/mshtml.py
+++ b/webview/platforms/mshtml.py
@@ -40,14 +40,14 @@ from WebBrowserInterop import IWebBrowserInterop, WebBrowserEx
 logger = logging.getLogger('pywebview')
 
 settings = {}
-    
+
 class MSHTML:
     class JSBridge(IWebBrowserInterop):
         __namespace__ = 'MSHTML.JSBridge'
         window = None
 
         def call(self, func_name, param, value_id):
-            return js_bridge_call(self.window, func_name, param, value_id)
+            return js_bridge_call(self.window, func_name, json.loads(param), value_id)
 
         def alert(self, message):
             BrowserView.alert(message)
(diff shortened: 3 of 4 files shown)


r0x0r:
Sorry for the late response @peter23 
I tested this briefly on Windows, `js_api.py` example is broken with MSHTML. `drag_region.py` still throws an error too.
EdgeHTML and Chromium are fine.

peter23:
@r0x0r Could you re-check please? Both examples works fine on my end.

And changes are quite small, I can not even imagine some problem.

`pywebview._bridge.call` is called only from `js/api.py` and `js/drag.py`

Now it is always called without any encoding.

`pywebview._bridge.call` sends a message to renderer, uses `JSON.stringify` of parameters in all cases for any renderer.
('chromium' does that automatically as I understand)

All renderers use `json.loads`.
So no need to `json.loads` in `webview/util.py`.

And that's all.
Do not see anything suspicious.

r0x0r:
I have tested it again and MSHTML indeed fails. js_api.py does not display any results (no errors too). Have you set gui='mshtml'?

peter23:
I re-run tests twice, I re-run them on several machines - works fine.
I can only guess that somehow you use old pywebview code when you run tests.

I created a clean virtual machine with Windows 7, installed IE 11, .Net Framework 4.8, Python 3.7.9, Git and made a video demo:
https://screencast-o-matic.com/watch/crVelPPf4b

I use `sys.path.insert(0, '...')` to use pywebview from specified folder instead of installed with pip.

peter23:
Virtual machine itself (for virtualbox) if you want: https://mega.nz/file/F0QCGRDS#5cTsBycqwGY4lhJEG8kXCSka2nxZmoly0yy9T2-BmhM

r0x0r:
This is getting more weird. I spent over an hour investigating this issue and 

1) js_api example works sometimes but fails on other occassions. The error is this
image
It appears like a timing issue, as my vm is extremely slow

2) could not get drag_region to work at all.

r0x0r:
Ok both problems fixed.
The BrowserWindow error was caused by a debug alert. Subsequently alert was broken with MSHTML.

drag_region was caused by MSHTML sending window coordinates in float and SetWindowPos expecting them in int.

I pushed the fixes to master.

peter23:
@r0x0r thank you for finding and fixing these bugs, BUT these bugs are not related with all three issues which I described and fixed. So, all three issues are still there. Basically my changes are the only logical way to fix them, so you could just merge them or I could add some details if you have some doubts.

One more time about all three issues:

1. Array of `x` and `y` is transferred from 'mshtml' to python without json-encoding.
That array become 'System.__ComObject' in python.
That object can not be used as 'tuple' or 'list', so it can not be used as `*param`, so `window.move(*param)` fails.
So it is not about int or float, the whole array is lost.
The solution is obvious and clean: use encoding.

2. JS-code for 'qtwebkit' also does not use json-encoding.
And JS-code for 'qtwebengine' uses json-encoding.
But `platforms/qt.py` is used for both platforms on python side. And it expects json-encoded params.
So 'qtwebkit' is broken.
Solution is the same: use encoding for 'qtwebkit'.

3. Now if we always use encoding in `window.pywebview._bridge.call` then we do not need double json-encoding, we can remove encoding from `window.pywebview.api`.

peter23:
Video demo that drag_region fails with the same error:

https://user-images.githubusercontent.com/2797301/103922413-60ce5080-511c-11eb-9621-f1995f2bb67b.mp4

Issue #531, opened 2020-05-26
Start clean instance of webview


bvukotic:
### Specification
- pywebview version: 3.2
- platform / version: Any

### Description  
  
It would be useful to start "clean" instance of pywebview (like incognito mode). Currently, cookies are remembered somewhere in the web engine, so each time I start webview session is not clean, some logins are remembered, etc... 
It would be nice to have something like 

webview.start(clean=True)

JavaFX  WebView starts clean every time (by default)

r0x0r:
Good suggestion. Clean state could be set default.

r0x0r:
API specs 
`webview.start((clean_session=True)` (enabled by default)

Links to implementations 

Cocoa: https://stackoverflow.com/questions/42994553/creating-a-non-tracking-in-app-web-browser

GTK: https://webkitgtk.org/reference/webkit2gtk/unstable/WebKitCookieManager.html
https://stackoverflow.com/questions/12220373/webkit-is-it-possible-to-store-cookies-to-file-and-reuse-it-again/15800193

QT: https://doc.qt.io/qt-5/qml-qtwebengine-webengineprofile.html

CEF: https://magpcss.org/ceforum/apidocs3/projects/(default)/CefRequestContext.html

MSHTML: https://stackoverflow.com/questions/13793261/disable-cookie-read-write-in-webbrowser-c-sharp-application

EdgeHTML: ¯\\\_(ツ)\_/¯
Cookies are shared only if launched in the same host process https://docs.microsoft.com/en-us/windows/communitytoolkit/controls/wpf-winforms/webview

bvukotic:
Can I do this in my python code? 
I am on windows 10, I guess I (we) use mshtml, can I call the code mentioned here
https://stackoverflow.com/questions/13793261/disable-cookie-read-write-in-webbrowser-c-sharp-application

but form python?

r0x0r:
This can be implemented, but it might require C# code. WebBrowserInterop.dll acts like a bridge for things that cannot be implemented in Python.

analogaio:
This would be a great addition to webview

r0x0r:
@Ksengine if you want to help, this issue would be a great choice.

Ksengine:
I can do qt and gtk.

obfusk:
In my experience, the gtk webview doesn't store cookies.
I opened #687 to not save cookies in qt webengine.

Ksengine:
[...]

PR #687, opened 2021-01-18
QtWebEngine: use OTR profile

obfusk:
Use an off-the-record QWebEngineProfile, instead of (unexpectedly) persisting cookies.

Related: #531

### DIFF
--- a/webview/__init__.py
+++ b/webview/__init__.py
@@ -60,11 +60,12 @@ _debug = False
 _user_agent = None
 _multiprocessing = False
 _http_server = False
+_incognito = True
 
 token = _token
 windows = []
 
-def start(func=None, args=None, localization={}, gui=None, debug=False, http_server=False, user_agent=None):
+def start(func=None, args=None, localization={}, gui=None, debug=False, http_server=False, user_agent=None, incognito=True):
     """
     Start a GUI loop and display previously created windows. This function must
     be called from a main thread.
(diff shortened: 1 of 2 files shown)


r0x0r:
Could you add a parameter to `webview.start` as well, so user could choose whether to save cookies or not? In the ticket I specified a `clean` parameter, but on the second thoughts `incognito` would be better. The default signature should be `webview.start(incognito=True)`
It can be implemented in a similar manner as debug parameter.

obfusk:
> Could you add a parameter to `webview.start` as well, so user could choose whether to save cookies or not? In the ticket I specified a `clean` parameter, but on the second thoughts `incognito` would be better. The default signature should be `webview.start(incognito=True)`
> It can be implemented in a similar manner as debug parameter.

Sure. The reason I didn't do that yet is that `qtwebkit` doesn't seem to have an easy way to have a *non*-incognito mode. So when `qtwebview` is not available I'd either have to error/warn/ignore when `incognito=False`. Or figure out a way to make it work.

How do you want to handle that?

obfusk:
Also:

Where should it store the data by default? Currently `qtwebengine` uses a global shared profile (`~/.config/QtWebEngine/Default` on Linux iirc).

It seems only cookies are supported, not e.g. `localStorage` :(

I think it might be best to just make everything incognito for now. But maybe provide a hook at the appropriate place so a user can easily add their own logic.

r0x0r:
logger.warning sounds like a plan for QtWebKit. Default location can be used as well.

Considering non-incognito mode is rather sketchy, it could be added as an experimental feature now and details can be ironed later on.

obfusk:
I added the parameter and the warning.

Issue #702, opened 2021-03-02
Request: Cross-platform access to screen frame rectangle/size


shepazon:
### Specification
- pywebview version: 3.4
- operating system: macOS
- web renderer: WebKit

### Description  
  
In order to aid in positioning windows, it would be useful to have information about the geometry of the screen (or, better, all the screens, if there are multiple, but I'd settle for just the "main" screen). For just the main screen, just width and height is enough, but for multiple you'd need the frame rectangle of each display in order to understand how to place the window where you want it on a given screen.

My use case: I want to position my window at the top-left corner of the main screen. I can do that for Mac by using PyObjC, but then I'm not able to run on Linux or Windows, which I'm trying to be able to do. Since PyWebView already deals with the various platforms' runtimes, it should be able to relatively easily get this information and expose it.

For just width and height of the main screen, could be done using a `webview.screen_size` tuple. For multiple screens, `webview.screens[0]` being the main screen, and subsequent objects for added screens. Each of these would be either a tuple or a rectangle object of some form. Wonderful :)
  
### Practicalities

- YES/NO I am willing to work on this issue myself.

Unfortunately, I can't at this time due to time constraints.

- YES/NO I am prepared to support this issue financially.

Yep, I think that could be arranged.

shepazon:
Actually, another option would be to simply make it so you could use negative numbers for the `x` and `y` parameters to specify the position of the window relative to the bottom-right of the screen instead of the top-left. Or mix and match. Like (40, -40) to put the window 40 pixels from the left edge and 40 from the bottom.

Then you don't need to do any fussing with adding a bunch of new API for getting screen sizes, at least for the use case of wanting to position a window relative to the other edges of the screen.

There are a lot of interesting things you could do going beyond this, like accepting percentages instead of pixel values for position and size information, but that's another issue perhaps. :)

(FWIW, even if things like this won't come to BitBar but I hope they get thought about for the new project)

r0x0r:
Good suggestion. I would go with `webview.screens` that would contain a list of tuples (width, height). screens[0] would always point to the main screen (if applicable). I am not fond of using negative numbers, at least not at this point. 

[...]

Issue #704, opened 2021-03-03
`y` parameter doesn't seem to work (macOS)


shepazon:
### Specification
- pywebview version: 3.4
- operating system: macOS 10.15.7
- web renderer: WebKit

### Description  
  The `y` parameter does not seem to be functional. I've tried setting it to all kinds of values, and nothing works. `x` works just fine. It doesn't matter whether I set it using `webview.create_window()` or `window.move()`. It just sits at its default vertical position.

  The only way not to have the window centered vertically is to set `x` to a specific value when creating the window. Then `y` is changed to be a short distance below the menu bar on my Mac. But I still can't choose a specific value of `y`.

  Looking through the code I also just discovered you can't set just one of `x` or `y` in `create_window()`; it would be fantastic to be able to just set `y` so it would be centered horizontally but, say, positioned near the top of the screen, or whatever. Anyway...

### Practicalities

- YES/NO I am willing to work on this issue myself.
Hypothetically though I've already looked through the code and it sure looks like it ought to "just work." Any amount of complexity and it's probably beyond the scope of the time I have.

- YES/NO I am prepared to support this issue financially.

Yes, within reason. :)

r0x0r:
I have just tested with `window = webview.create_window('Simple browser', 'https://pywebview.flowrl.com/hello', x=0, y=200)
` and `window = webview.create_window('Simple browser', 'https://pywebview.flowrl.com/hello', x=0, y=0)`. Both cases work as expected.

Tested with a MacBook with no external monitor attached. What is your setup?

shepazon:
I do have an external monitor as a secondary display.

r0x0r:
This has to do with reprojecting Cocoa coordinates (Cartesian with 0,0 at bottom-left) with computer ones (0,0 at top-left). See https://github.com/r0x0r/pywebview/blob/3a45089c85bf4c675c7488a7fc4251bd61e79ca4/webview/platforms/cocoa.py#L501

This approach seems to work with a single monitor, but in a multi-monitor setup produces weird results. I am not sure how to tackle this. Ideas are welcome.

shepazon:
This may be something that has to be addressed in the context of the `webview.screens` array of screen information objects; if a window has both a `y` property and a `window.screen` which is a reference to one of the screens in `webview.screens`, you can then use the size information found in that along with the value of `y` to compute the actual position in top-down fashion.

--
Eric Shepherd
Sr. Developer Writer
AWS Documentation

[...]

PR #711, opened 2021-03-22
webview.screens

r0x0r:
https://github.com/r0x0r/pywebview/issues/702

Implement `webview.screens` property.

This PR introduces a new dependency.
`pip install proxy_tools`

### DIFF
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -492,11 +492,10 @@ class BrowserView:
         self.window.deminiaturize_(self)
 
     def move(self, x, y):
-        screen_frame = AppKit.NSScreen.mainScreen().frame()
-        if screen_frame is None:
-            raise RuntimeError('Failed to obtain screen')
+        screen = get_screens()[0]
 
-        flipped_y = screen_frame.size.height - y
+
+        flipped_y = screen.height - y
         self.window.setFrameTopLeftPoint_(AppKit.NSPoint(x, flipped_y))
 
     def get_current_url(self):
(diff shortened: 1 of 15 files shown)


shepazon:
This is working for me, as advertised. It isn't particularly useful in an environment with multiple displays (of different heights, anyway) with the bug in issue #704 unresolved... but this does seem to be working well. Probably needs to be tested on a wide variety of configurations, of course.

webview/platforms/cocoa.py
```diff
@@ -492,11 +492,10 @@ def restore(self):
         self.window.deminiaturize_(self)
 
     def move(self, x, y):
-        screen_frame = AppKit.NSScreen.mainScreen().frame()
```
shepazon, line 496: How about using `self.window.screen().frame()` here instead? This guarantees getting the correct window's frame instead of the frame of the active window, which might not be the one being moved.

shepazon (approved):
Even with this change, I'm still not getting functional positioning of my windows. No matter what I set `y` to in `webview.create_window()`, I still get a window flush up against the top of the screen, nestled against the menu bar.

shepazon (changes requested):
Meant to request changes, not approve. :)

shepazon (approved):
OK, I had a configuration issue here. It actually does work, though my suggestion about using `NSWindow.screen` to get the window's screen stands as a more reliable solution that's less likely to accidentally get the wrong screen in some situations.

r0x0r:
NSWindow.screen() somewhat works. It returns a correct screen after window is opened, but keyboard focus screen just before the launch. It is certainly better than mainScreen(), but `move_window.py` still produces incorrect results.

Issue #697, opened 2021-02-24
Per-window localization


Fizzadar:
### Description  
  
Customize the `localization` dictionary on a per-window basis; for example a custom "You have unsaved changes, are you sure you wish to close the window" kinda thing specifically for a settings window. Currently can easily override the defaults but it'd be nice to have per-window API support. Can look into this myself if pro the idea :)
  
### Practicalities

- YES I am willing to work on this issue myself.
- NO I am prepared to support this issue financially.

r0x0r:
I don't see a reason why not. Should be a `localization` parameter to `create_window` with its taking precedence over the global localization?

Fizzadar:
Also  thought the same, hoping to get a chance to look at this over the weekend!

Fizzadar:
Not stale, just not had any time to look at this yet but is on my todo list!

Fizzadar:
Not stale, draft PR in progress now! https://github.com/r0x0r/pywebview/pull/726

PR #726, opened 2021-05-12
Localization per window

Fizzadar:
This is a bit fiddly due to the `create_window` calls happening before the `webview.start` call, hence the need to update the localization object within the `_initialize` function.

This maintains support for the existing API, the alternative is to remove `localization` from `webview.start` entirely, but this would be a breaking change. Keen to hear your thoughts before proceeding with the other GUI frameworks!

Closes: #697

### DIFF
--- a/examples/localization.py
+++ b/examples/localization.py
@@ -24,6 +24,13 @@ if __name__ == '__main__':
         'linux.openFolder': u'Открыть папку',
     }
 
-    webview.create_window('Localization Example', 'https://pywebview.flowrl.com/hello')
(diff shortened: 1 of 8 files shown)


Fizzadar:
Not stale! @r0x0r do you have any thoughts on this? If you're happy with the approach I can apply it to the other GUI libs.

r0x0r:
@Fizzadar Yes please go ahead with other platforms. This looks fine to me.

Fizzadar:
@r0x0r should be ready to go now, remaining platforms implemented!

PR #747, opened 2021-07-11
debugging note

Mohamed501258:
This small note is to demonstrate that, the DevTools through the context menu will not open unless two parameters are included, the debug=True, and gui="cef"

### DIFF
--- a/docs/examples/debug.md
+++ b/docs/examples/debug.md
@@ -1,7 +1,7 @@
 # Debugging
 
 To open up debugging console, right click on an element and select Inspect.
-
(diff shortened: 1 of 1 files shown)


r0x0r:
This is no good, as CEF is only one of many available renderers. For Windows, devtools is available for CEF and EdgeChromium.

r0x0r:
DevTools on right click is available both for CEF and EdgeChromium. EdgeHTML supports devtools too via an external app you have to download separately. Only MSHTML lacks any kind of debugging tools.

Mohamed501258:
Is it possible to add drag-drop files feature to pywebview and get the file
full path?

Mohamed501258:
When I try to include js module inside the html file, and run it using
pywebview it gives"access not allowed"
Also html video doesn't work on pywebview

How to solve these things

PR #751, opened 2021-07-22
Solving: file access not allowed on "cef" renderer 

Mohamed501258:
using "cef" renderer will prevent the local files access, this edit will allow serving local files from different folders without problems.

### DIFF
--- a/webview/platforms/cef.py
+++ b/webview/platforms/cef.py
@@ -236,11 +236,14 @@ def init(window):
 
         _initialized = True
 
+browser_settings = {
+    "web_security_disabled": True,
+}
 
 def create_browser(window, handle, alert_func):
     def _create():
         real_url = 'data:text/html,{0}'.format(window.html) if window.html else window.real_url or 'data:text/html,{0}'.format(default_html)
-        cef_browser = cef.CreateBrowserSync(window_info=window_info, url=real_url)
+        cef_browser = cef.CreateBrowserSync(window_info=window_info, url=real_url, settings=browser_settings)
         browser = Browser(window, handle, cef_browser)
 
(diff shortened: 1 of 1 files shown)


r0x0r:
Are there any other ramifications of disabling web security?

Mohamed501258:
as we aim to create desktop applications that have full access to files on
the pc, I think it will not be problem

r0x0r:
Disabling web security also disables CORS, SSL checks and OPTIONS. I don't think it would be a wise thing to disable it by default. Instead you could disable web security in your application like this.

```
from webview.platforms.cef import settings
settings.update({
    'disable_web_security': True
})
```

Mohamed501258:
But this issue limits the access to files outside the working directory,
and of course when there is relatively big project it will be a problem.I
agree it is not safe, but is there any alternative solution for this? It is
supposed to be desktop application with full access to the files otherwise
it will affect the performance.
This is my opinion

بتاريخ الأحد، 1 أغسطس 2021، جاء من Roman ***@***.***>:

r0x0r:
I agree that this is indeed an issue, but disabling web security is too of a drastic measure. I propose documenting the approach I had posted above.

r0x0r:
CEF seems to have allow-file-access-from-files switch that is exactly what is needed

Mohamed501258:
Yes, I agree. But the pros and cons of this modification should be
highlighted in the doc so who writes the code is aware of the security
issues

Mohamed501258:
Yes it has, and I have tried this before but the file access doesn't work
unless I disable web security.


بتاريخ الثلاثاء، 3 أغسطس 2021، جاء من Roman ***@***.***>:

PR #763, opened 2021-08-13
[WIP] Allow fixing window position to any side

Timoses:
Implements ability to configure which edge fixes the windows position.

**Only Cocoa implemented so far**.

I had a quick look at Gtk which uses [Gravity](https://lazka.github.io/pgi-docs/#Gdk-3.0/enums.html#Gdk.Gravity).
Not sure what the best approach is on pywebview side to implement it for all platforms.

Further, it should be possible to dynamically adjust the gravity point / fix position of the window during runtime.
Hence, I believe there should be a method available on `webview.window.Window` to set it.

Do you have any valuable insight and suggestions?


r0x0r:
Sorry for the late response. I have tested this with 
```
    window = webview.create_window('Simple browser', 'https://pywebview.flowrl.com', fix_point=webview.FixWindow.SOUTH)
    webview.start()
```
and could not see any difference. How should it behave?

Timoses:
> Sorry for the late response. I have tested this with
> 
> ```
>     window = webview.create_window('Simple browser', 'https://pywebview.flowrl.com', fix_point=webview.FixWindow.SOUTH)
>     webview.start()
> ```
> 
> and could not see any difference. How should it behave?

Ty for taking a look.

The effect should come into play when you dynamically resize the window (https://github.com/r0x0r/pywebview/blob/master/webview/window.py#L240).

Default is that the top left corner of the window stays on the same position. There are however scenarios when other sides should stay fixed.

An example would be a notification window with variable size which changes over time. If the window were positioned on the right side of the screen and grew then in the current setup it would potentially grow over the right side of the screen. With this PR it can be set to be fixed on the 'EAST' side so that the window grows towards the left side.

Timoses:
Example - should give you a steadily growing window which you can move around (not tested):

```
def adjust_loop(window, screen_w_2, screen_h_2):
    x = 0
    y = 0
    width = 0
    height = 0

    while True:
        time.sleep(0.5)
        if window.x < screen_w_2:
            window.load_css('p { text-align: left; }')
            # Currently in PR https://github.com/r0x0r/pywebview/pull/763
            if hasattr('window', 'fix_point'):
                window.fix_point = window.fix_point & (~ webview.FixWindow.EAST)
        else:
            window.load_css('p { text-align: right; }')
            # Currently in PR https://github.com/r0x0r/pywebview/pull/763
            if hasattr('window', 'fix_point'):
                window.fix_point = window.fix_point | webview.FixWindow.EAST

        if window.y < screen_h_2:
            # Currently in PR https://github.com/r0x0r/pywebview/pull/763
            if hasattr('window', 'fix_point'):
                window.fix_point = window.fix_point & (~ webview.FixWindow.SOUTH)
        else:
            # Currently in PR https://github.com/r0x0r/pywebview/pull/763
            if hasattr('window', 'fix_point'):
                window.fix_point = window.fix_point | webview.FixWindow.SOUTH

        width = width + 1
        height = height + 1
        window.resize(width+10, height+10)

def start_me():
    # Not sure, pywebview does not seem to like it when target URL is not
    # reachable yet
    time.sleep(10)
    window = webview.create_window('Woah dude!', url=https://github.com/r0x0r/pywebview/pull/763, on_top=True,
                                   frameless=True, width=500, height=500,
                                   transparent=False, resizable=False)
    screen_w = webview.screens[0].width / 2
    screen_h = webview.screens[0].height / 2
    webview.start(adjust_loop, (window, screen_w, screen_h), debug=True)

```

r0x0r:
Thanks for the explanation. I see what this is all about now.
What do other platforms look like? Would you be able to implement this feature for all the platforms?
If no native support for this feature is available on other platforms, it can be mimicked by adjusting both dimensions and position of the window.

Another suggestion is it would be better to have this as an additional optional parameter for to the resize method, eg. window.resize(100, 200, FixPoint.SOUTH)

r0x0r:
The documentation for SetWindowPos can be found here
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowpos

Timoses:
Finally came around to an attempt to implement remaining platforms.

I did not test it. Also *cef* implementation is not complete.

 @r0x0r could you please have a look at it?

webview/platforms/cef.py
```diff
@@ -109,8 +109,25 @@ def initialize(self):
...
-        windll.user32.SetWindowPos(self.inner_hwnd, 0, 0, 0, width - 16, height - 38,
+    def resize(self, width, height, fix_point):
+
+        # How much to move to the right
+        move_x = 0
+        # How much to move to the bottom
+        move_y = 0
+
+        rect = windll.?.RECT()
+        rect = windll.user32.GetWindowRect(self.inner_hwnd, rect)
```
Timoses, line 120: @r0x0r How would I do this correctly? Do you know?

PR #812, opened 2021-11-17
Update README.md

thewh1teagle:
because pythonnet package we can use only python 3.8 or older in this package, hence I suggest adding it to the README instead of let the people figure out themself the problem and the solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ _pywebview_ is created by [Roman Sirokov](https://github.com/r0x0r/).
 
 ### Install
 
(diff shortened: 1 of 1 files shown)


r0x0r:
Good suggestion. Could you add that this limitation only applies to Windows?

thewh1teagle:
Added, 
Also, I suggest you to add it to the website of pywebview.
It's important, this bug almost made me think this library won't work because it took me a lot of time to figure out it.

PR #848, opened 2022-02-02
replace pyqt5 with qtpy

tshemeng:
1. replace pyqt5 with [qtpy](https://pypi.org/project/QtPy/) as abstraction layer in the code, people can freely choose pyqt5, pyside2, pyside6 as the backend
2. replace default pywebview[qt] backend from pyqt5(GPL) to pyside2(LGPL), the LGPL License is more commercial friendly
3. add qt engine support for Windows, people can distribute Qt's built-in Chromium engine, and without requires for users to install additional Webview2(100+MB)

for now [QtWebEngine/ChromiumVersions](https://wiki.qt.io/QtWebEngine/ChromiumVersions)

|PySide2&6|Qt Version|Chromium Version|
| ---------- | ---------- | -------------------|
|PySide6 6.2.3|6.2.0|90.0.4430.228|
|PySide2 5.15.2.1|5.15.2|83.0.4103.122|

### DIFF
--- a/docs/contributing/development.md
+++ b/docs/contributing/development.md
@@ -84,5 +84,8 @@ pytest tests/test_simple_browser.py
 
 ### Linux 
 * [PyGObject API reference](https://lazka.github.io/pgi-docs/)
-* [PyQt5 reference guide](http://pyqt.sourceforge.net/Docs/PyQt5/)
+
+### Qt
+* [Qt for Python Documentation](https://doc.qt.io/qtforpython-5/contents.html)
 * [Qt5 documentation](https://doc.qt.io/qt-5/index.html)
+* [PySide2 QtWidgets](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/index.html)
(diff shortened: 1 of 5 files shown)


r0x0r:
PySide support is greatly welcome.  Is QtPy fully compatible with PyQt5 though?
I have no capacity to thoroughly test it.

tshemeng:
QtPy use PyQt5 as default backend, the classes, functions imported from QtPy in fact imported from original PyQt5 modules, `isinstance` works as usual for downstream.

The breaking thing is I change `pip install pywebview[qt]` to install `pyside2`, so maybe we keep `pip install pywebview[qt]` install `pyqt5`, and add additional `pywebview[pyside2]`, `pywebview[pyside6]` to `setup.py`.

tshemeng:
Another incompatibility is that QtPy requires Python >= 3.6, and now pywebview works on Python 3.4+.

r0x0r:
Ok sounds good. I am fine with 3.6+

Issue #879, opened 2022-04-03
【bug】There is a problem with comparing webView versions when version is 100.0.xxx


greper:
![1111](https://user-images.githubusercontent.com/1687298/161431754-26acfaa6-2f51-4ba7-8ab0-4a274cce0921.png)

![2222](https://user-images.githubusercontent.com/1687298/161431745-a9257203-a662-4086-b111-8f7d61ad9b12.png)

dmwyatt:
Not to put a too fine a point on it, but this breaks Edge WebView2 completely if it is up-to-date.

kevinbevers:
This should be immediately pushed out to release. Its a breaking change when edge is up to date

THEGOLDENPRO:
Does this explain why I'm getting this error, even when attempting to run the "edgechromium" renderer?

**Output**
```
[pywebview] [WinError 2] The system cannot find the file specified - Microsoft Edge WebView2 Beta Registry path: Computer\HKEY_CURRENT_USER\Microsoft\EdgeUpdate\Clients\{2CD8A007-E189-409D-A2C8-9AF4EF3C72AA}
[pywebview] [WinError 2] The system cannot find the file specified - Microsoft Edge WebView2 Beta Registry path: Computer\HKEY_LOCAL_MACHINE\WOW6432Node\Microsoft\EdgeUpdate\Clients\{2CD8A007-E189-409D-A2C8-9AF4EF3C72AA}
[pywebview] [WinError 2] The system cannot find the file specified - Microsoft Edge WebView2 Developer Registry path: Computer\HKEY_CURRENT_USER\Microsoft\EdgeUpdate\Clients\{0D50BFEC-CD6A-4F9A-964C-C7416E3ACB10}
[pywebview] [WinError 2] The system cannot find the file specified - Microsoft Edge WebView2 Developer Registry path: Computer\HKEY_LOCAL_MACHINE\WOW6432Node\Microsoft\EdgeUpdate\Clients\{0D50BFEC-CD6A-4F9A-964C-C7416E3ACB10}       
[pywebview] [WinError 2] The system cannot find the file specified - Microsoft Edge WebView2 Canary Registry path: Computer\HKEY_CURRENT_USER\Microsoft\EdgeUpdate\Clients\{65C35B14-6C1D-4122-AC46-7148CC9D6497}
[...]

PR #880, opened 2022-04-03
fix: Fix the problem with comparing webView versions when version is 100.0.xxx

greper:
https://github.com/r0x0r/pywebview/issues/879


dmwyatt:
This works in my testing.  👍👍

r0x0r:
Thanks for the fix. I will release a new version soon.

Issue #500, opened 2020-04-14
Bar menus


bustawin:
### Description  
Although this question is not strictly about pywebview if I can get some help I could write a guide with the code or even work with a PR in the future if there is interest.

I have seen that a menu bar is quite required for my app (ex. File > Open, File > Save, Edit > ...). Although I plan to release it in the three platforms, I am starting with Mac, which let me try to hijack the Cocoa's BrowserView._add_view_menu to add simple menu items that would ultimately just call a custom python function.

I have never touched PyObjC. Just mimicking the existing code I can create new menus
`myMenu.addItemWithTitle_action_keyEquivalent_("full", "foobar:", "o")` but I don't know how to do it to execute my python function (I understand I need to add a foobar_() function somewhere...). PyObjC guides / example code have not helped me, as there are differences between the ones I have seen and your implementation. Could I get some help?

I understand in Linux / Windows this task will less cryptic?

Thank you!

Btw, amazing tool you have here :-)
  
### Practicalities

- YES I am willing to work on this issue myself.
- NO I am prepared to support this issue financially.

r0x0r:
Objective C is confusing af and can be intimidating. PyObjC adds its own quirks, but the good news is that it mostly works, once you get the syntax. As for the menu, somebody else has implemented it, so myself I have only a vague idea how it should be done. `addItemWithTitle_action_keyEquivalent_` seems to be the right method and action is attached via a selector.  Here is an example of how to use a selector in Python. https://lethain.com/how-to-use-selectors-in-pyobjc/


The menubar has been requested several times and in my opinion it can be incorporated into `pywebview`. If you would like to extend your work into a generic solution, that would be most welcomed. The scaffold for the API would be something like this 
```
menu_items = {
  'File': (
    ('First item', action),
    (), // separator
    ('Second item', action2),
  )
  'Edit': ( ....)
}
menu = webview.create_menu(menu_items)
window = webview.create_window('Window with menu', menu=menu)
```

bustawin:
Thank you! I will take a look.

I will see if I can make something generic work for my project and come back.

bustawin:
I have been able to create a working menu :-)

Now the problem is that I don't know how to relate the menu to the active window. For now, the menu is global. So if I click "save" and I have several windows I don't know which window to save.
Ideally the `my_action()` should receive a `window` argument to know which window executed it.
 Any idea how to get it?

```python
from __future__ import annotations

import typing as t

import AppKit
import objc
[...]

PR #886, opened 2022-04-23
Application Menubar

sardination:
Implements a bar menu functionality. On Cocoa and supported Qt and GTK platforms, this manifests as a bar menu at the top of the screen. On all other platforms, this manifests as a menubar at the top of each window of the application.

This also implements the `webview.active_window` function that returns the active pywebview window at the time of calling.

See `examples/app_menu.py` for an implementation example.

Notes:
* Addresses and borrows a code snippet from the thread for #500
* Adjusts some basic application logic by abstracting out a `setup_app` function and also changing how GTK starts an application.


r0x0r:
Great work and many thanks for tackling this. I will review and test this asap.

webview/__init__.py
```diff
@@ -67,8 +68,10 @@
 
 token = _token
 windows = []
+menus = []
 
-def start(func=None, args=None, localization={}, gui=None, debug=False, http_server=False, user_agent=None):
+def start(func=None, args=None, localization={}, gui=None, debug=False, http_server=False, user_agent=None,
+          app_menu_list=[]):
```
r0x0r, line 74: `app_menu_list=` could be just `menu=` for clarity. Also would it be possible to make a menu window not app specific? Ideally both `create_window` and `start` should accept `menu` param. The `create_window` menu would a be window specific. If both are specified, the `create_window` menu would replace the global menu.

webview/__init__.py
```diff
@@ -120,6 +124,7 @@ def _create_children(other_windows):
         raise WebViewException('You must create a window first before calling this function.')
 
     guilib = initialize(gui)
+    guilib.setup_app()
```
r0x0r, line 127: Could you combine `setup_app` this into `initialize(gui)`

webview/platforms/gtk.py
```diff
@@ -458,14 +459,27 @@ def create_bridge():
...
     def create():
         browser = BrowserView(window)
         browser.show()
 
-    if window.uid == 'master':
+    def create_master_callback(app):
         create()
+
+    if window.uid == 'master':
+        _app.connect("activate", create_master_callback)
```
r0x0r, line 479: The project uses single quotes.

r0x0r (changes requested):
Thanks again for this massive effort. Please change double quotes to single ones (apart from multiline strings) and `change the app_menu_list` to `menu`. Window-specific menu can be postponed for later if it is too much work.

r0x0r:
Cannot run `app_menu.py` with GTK, the error is 
```
(app_menu.py:37056): GLib-GIO-ERROR **: 11:21:51.525: g_menu_item_set_detailed_action: Detailed action name 'app._Test Menu_Change Active Window Content' has invalid format
```

r0x0r:
On Windows menu appears only in the second window.

image

webview/__init__.py
sardination: I spent some time trying to get window-specific menu working on all platforms and I haven't been able to get it to work fully, so I'm going to leave that out of this PR.

sardination:
I believe I fixed all of the noted issues, but haven't gotten around to testing it one more time on all platforms. I tried to see if I could add the per-window menu feature quickly enough, but it wasn't behaving well on all platforms, so I kept this PR to just be for an app-wide menu.

sardination:
> I am still getting GTK error. What pygobject / GTK versions are you using?
> 
> Winforms cannot start too.

@r0x0r sorry I didn't get around to this - I'm glad it seems like you've addressed the problems. Once I have access to a Windows computer again I will test functionality.

Issue #906, opened 2022-06-21
Async evaluate_js results in exception in background thread


detritophage:
### Specification
- pywebview version: 3.6.3
- operating system: Windows
- web renderer: Chromium

### Description  
  
Async version of `evaluate_js` causes exception in background thread.

Using a slightly modified version of this code: [evaluate_js_async.py](https://github.com/r0x0r/pywebview/blob/master/examples/evaluate_js_async.py) (see updated lines `20-25`)

```python
01   import webview
02   
03   """
04   This example demonstrates evaluating async JavaScript
05   """
06   
07   def callback(result):
08       print(result)
09   
10   def evaluate_js_async(window):
11       window.evaluate_js(
12           """
13           new Promise((resolve, reject) => {
14               setTimeout(() => {
15                   resolve('Whaddup!');
16               }, 300);
17           });
18           """, callback)
19   
20 + def test():
21 +     print("JS called Python")
22 + 
23   if __name__ == '__main__':
24 +     window = webview.create_window('Run custom JavaScript', html='<html><body><div onclick="pywebview.api.test()">TEST</div></body></html>')
25 +     window.expose(test)
26       webview.start(evaluate_js_async, window, debug=True)
```

Run the above script, wait for "Whaddup!" to print, and click the `TEST` text. Something like the following output will be observed:
```
JS called Python
Exception in thread Thread-1:
Traceback (most recent call last):
  File "C:\Applications\Development\Python38\lib\threading.py", line 932, in _bootstrap_inner
    self.run()
  File "C:\Applications\Development\Python38\lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Data\Development\test\asyncExample.py", line 13, in evaluate_js_async
    window.evaluate_js(
  File "C:\Data\Development\test\lib\site-packages\webview\window.py", line 34, in wrapper
    return function(*args, **kwargs)
  File "C:\Data\Development\test\lib\site-packages\webview\window.py", line 378, in evaluate_js
    return self.gui.evaluate_js(escaped_script, self.uid)
  File "C:\Data\Development\test\lib\site-packages\webview\platforms\winforms.py", line 683, in evaluate_js
    return BrowserView.instances[uid].evaluate_js(script)
  File "C:\Data\Development\test\lib\site-packages\webview\platforms\winforms.py", line 321, in evaluate_js
    result = self.browser.js_results[id]
KeyError: '322286d1'
```

[...]

Issue #905, opened 2022-06-21
Async evaluate_js crashes application


detritophage:
### Specification
- pywebview version: 3.6.3
- operating system: MacOS
- web renderer: WebKit

### Description  
  
Async version of `evaluate_js` crashes the app.

Using this code: [evaluate_js_async.py](https://github.com/r0x0r/pywebview/blob/master/examples/evaluate_js_async.py)
  
Run the above script and the application will crash with something like the following:

```
[pywebview] Using Cocoa
2022-06-08 15:00:28.279 Python[23737:181246] *** Terminating app due to uncaught exception 'OC_PythonException', reason: '<class 'TypeError'>: the JSON object must be str, bytes or bytearray, not bool'
*** First throw call stack:
(
        0   CoreFoundation                      0x00000001a4c6eb08 __exceptionPreprocess + 240
        1   libobjc.A.dylib                     0x00000001a49b9e14 objc_exception_throw + 60
        2   _objc.cpython-39-darwin.so          0x0000000103773dd4 PyObjCUtil_Strdup + 0
        3   _objc.cpython-39-darwin.so          0x000000010375720c method_stub + 6072
        4   libffi.dylib                        0x00000001b2d94f34 ffi_closure_SYSV_inner + 816
        5   libffi.dylib                        0x00000001b2d8c1e8 ffi_closure_SYSV + 56
        6   WebKit                              0x00000001c3784d64 _ZN3WTF6Detail15CallableWrapperIZN6WebKit12WebPageProxy33runJavaScriptInFrameInScriptWorldEON7WebCore23RunJavaScriptParametersENSt3__18optionalINS_16ObjectIdentifierINS4_19FrameIdentifierTypeEEEEERN3API12ContentWorldEONS_17CompletionHandlerIFvONSt12experimental15fundamentals_v38expectedINS_6RefPtrINSD_21SerializedScriptValueENS_12RawPtrTraitsISL_EENS_21DefaultRefDerefTraitsISL_EEEENS4_16ExceptionDetailsEEEEEEE4$_22vJONS_4SpanIKhLm18446744073709551615EEEONS8_ISR_EEEE4callES11_S13_ + 432
[...]

Issue #904, opened 2022-06-21
Async evaluate_js prevents process termination


detritophage:
### Specification
- pywebview version: 3.6.3
- operating system: Windows
- web renderer: Chromium

### Description  
  
Async version of `evaluate_js` prevents process termination.

Using this code: [evaluate_js_async.py](https://github.com/r0x0r/pywebview/blob/master/examples/evaluate_js_async.py)

Run the above script, wait for "Whaddup!" to print, and close the application window. Input control will not be returned in the command line terminal without killing the Python process in Task Manager.

If I remove the `callback` argument from `evaluate_js`, then closing the application window also terminates the process.
  
### Practicalities

- YES I am willing to work on this issue myself.

- NO I am prepared to support this issue financially.

PR #908, opened 2022-06-25
#904 #905 #906 Fix callback return value in async evaluate_js

detritophage:
Javascript generated for asynchronous evaluate_js call returned an actual boolean instead of a JSON stringified boolean.


r0x0r:
I could not reproduce #905 and #906, but this fix at least patches #904. Thanks for the contribution.

PR #924, opened 2022-07-29
fix evaluate invalid js for cef

irtimir:
if a script is executed with an error, then further script execution gets stuck


irtimir:
@r0x0r, do you think it is possible to merge these changes?

r0x0r:
Sure, sorry for the delays.

PR #942, opened 2022-08-27
Run menu functions on thread

sardination:
Certain functions would freeze the app when called from the menu because they were running on the main thread. Instead, call the functions from a thread.

Has been tested on Mac, needs to be tested on Linux and Windows. Can be tested by running `examples/menu.py` and selecting the menu option at *Test Menu > Random > File Dialog*. This should show a file dialog without freezing.

### DIFF
--- a/webview/platforms/qt.py
+++ b/webview/platforms/qt.py
@@ -13,7 +13,7 @@ import webbrowser
 import socket
 from uuid import uuid1
 from copy import deepcopy
-from threading import Semaphore, Event
+from threading import Semaphore, Event, Thread
 import typing as t
 
 from webview import _debug, _user_agent, _private_mode, _storage_path, OPEN_DIALOG, FOLDER_DIALOG, SAVE_DIALOG, windows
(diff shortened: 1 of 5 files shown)


r0x0r:
@sardination This is a welcomed addition. What is the situation with QT?

sardination:
> @sardination This is a welcomed addition. What is the situation with QT?

Sorry! Just forgot about QT. Added the implementation - but I haven't been able to test anything other than Cocoa for now.

sardination:
Commenting to keep PR from being closed - needs tests on non-Cocoa platforms (which I can do next month).

webview/platforms/qt.py
```diff
@@ -726,7 +726,7 @@ def create_submenu(title, line_items, supermenu):
                 m.addSeparator()
             elif isinstance(menu_line_item, MenuAction):
                 new_action = QAction(menu_line_item.title)
-                new_action.triggered.connect(menu_line_item.function)
+                new_action.triggered.connect(Thread(target=menu_line_item.function).start)
```
r0x0r, line 729: This syntax does not work unfortunately. Nothing happens when menu item is clicked.

PR #949, opened 2022-09-11
Implement text dialog with ok/cancel buttons

sardination:
Add `create_text_dialog` window method that allows creation of a text dialog with custom title and message populated with "OK" and "Cancel" buttons. Returns `1` on "OK" and `0` on "Cancel". 

Only tested on Cocoa - needs testing on other platforms.


sardination:
Commenting to keep PR from being closed - needs tests on non-Cocoa platforms (which I can do next month).

r0x0r:
Sorry for the late response. Finally found time to test this.

- I have fixed GTK / QT implementations. Windows is working fine.
- 1 and 0 values should be replaced with True / False for better readability.
- `create_text_dialog` is too cryptic. `create_confirmation_dialog` could be a better name.
-  `window.confirm` on the js side could point to this function in a similar way as it is done with `window.alert`.

PR #962, opened 2022-10-02
localized key events

max-uho:
After noticing on my AZERTY keyboard that cmd+Q would not quit the app but cmd+A would, I modified platforms/cocoa.py to convert keys to keycodes.

I used the "keyboard" package to fetch the "key_to_scan_codes" function which adds a dependency.


max-uho:
(sorry the requirements.txt file was added to the pull request and is a bit messy. We just need to add the "keyboard" package without its dependencies)

max-uho:
I looked into it, but this simple function calls on these 4 imports. Would you want these 4 extra files to be added to the pull request?

Screenshot 2022-10-03 at 13 58 49

r0x0r:
No, that would not be ideal. I looked into these functions and they are rather short / trivial. The best solution would be to extract these functions into a single self-contained function.

max-uho:
Added a cocoa_keyboard.py file to the platforms folder and a couple of lines into cocoa.py.
Works as expected and doesn't need the "keyboard" dependency anymore.

r0x0r:
Thanks.
Would you mind adding license info on top of the cocoa_keyboard.py to be fully compliant?

max-uho:
Just added the license. Sorry, I'm super unfamiliar with the process, this is my first pull request. Let me know if that seems ok to you.

max-uho:
Upgraded to a M1 Pro laptop and noticed that the key events were broken. This reduces the the code drastically and still works on my older machine.

r0x0r:
Good job on getting rid of the keyboard dependency. I have tried to test your PR with the AZERTY layout, but for me it works both in the original and fixed version.

As for frameless, I am not a big fan of the buttonless parameter. But now thinking about this maybe buttonless frameless should be a default option as in other platforms.

max-uho:
@r0x0r The AZERTY layout is still causing problems with pywebview 4.0.1. 
Hitting cmd+A will quit the program, for example.
Any reason not to merge https://github.com/r0x0r/pywebview/pull/962/commits/64874b6127bb2d8aa13e1363b93e85e265f9844f ?

r0x0r:
@max-uho you closed the PR, so I assumed the problem was not relevant any more. Anyhow, would you remove buttonless parameter before merging?

Issue #433, opened 2019-11-24
External killing of the python process where main thread is running leaves zombie (Desktop App Web Viewer)


MirkoCovizzi:
### Specification
- pywebview version: 3.1
- platform / version: Python 3.7.5 x64 Windows 10, default web engine

### Description  

If the python process running the main script gets killed externally, one of the two Desktop App Web Viewer processes doesn't terminate, causing memory leak.
Steps to reproduce: execute any of the demos. 2 Desktop App Web Viewer processes will spawn. Kill the python process that runs the script. One of the two Desktop App Web Viewer processes (Microsoft.Win32WebViewHost) will stay alive.

MirkoCovizzi:
It appears the problem arises because the window.browser.js_result_semaphore is not released if BrowserForm.on_close() is not called from winforms.destroy_window(). Adding self.pywebview_window.browser.js_result_semaphore.release() at the end of BrowserForm.on_close() solves the issue.

r0x0r:
Thanks for investigating this. Could you create a PR?

MirkoCovizzi:
Sadly, the problem persists if the application terminates unexpectedly (because of exceptions or something else), because it doesn't have occasion to reach into the closing function. We need a way to safely release the semaphore from somewhere else that doesn't depend on the .net api.

r0x0r:
Might the atexit handler be of help?
https://docs.python.org/3/library/atexit.html

SebastienWae:
> Might the atexit handler be of help?
> https://docs.python.org/3/library/atexit.html

Unfortunately, that won't work.
> Note: The functions registered via this module are not called when the program is killed by a signal not handled by Python, when a Python fatal internal error is detected, or when os._exit() is called.

abmyii:
Sorry for butting in, but can the signal be caught?
https://docs.python.org/3/library/signal.html
https://nattster.wordpress.com/2013/06/05/catch-kill-signal-in-python/
https://stackabuse.com/handling-unix-signals-in-python/
https://www.devdungeon.com/content/python-catch-sigint-ctrl-c
https://stackoverflow.com/questions/18499497/how-to-process-sigterm-signal-gracefully

abmyii:
For Windows-specific kill signals:
https://stackoverflow.com/questions/47306805/signal-sigterm-not-received-by-subprocess-on-windows?rq=1
https://stackoverflow.com/questions/35772001/how-to-handle-the-signal-in-python-on-windows-machine
[...]

PR #987, opened 2022-11-21
[WinForms] Fix hang on window closing when JS evaluation is in progress

colingavin:
This PR fixes the issue reported at https://github.com/r0x0r/pywebview/issues/433#issuecomment-622522261 for WinForms. The JS result semaphore is released, and `None` is returned from `evaluate_js`. This is modeled on what is done for the GTK platform.


colingavin:
@r0x0r Thanks for the quick response!

Issue #12, opened 2015-10-26
Disable right-click context menus 


r0x0r:
Right clicking on the browser container generates a context menu on every platform. They must be disabled in order to hide implementation details and prevent unwanted actions such as page reload. Any input on how this could be achieved on any platform is appreciated.

jonathon-love:
under Qt, you can call

    webview->setContextMenuPolicy(Qt::NoContextMenu)

(under C++, not sure what the equivalent python would be) to disable the right click menu

r0x0r:
Implemented in master

PR #997, opened 2022-12-04
Avoid control-left-click context menu on MacOS, also allow right-click events to reach javascript handlers

ecpost:
Change 8c3b47b9533fe9d5980df15fcf2596603cfc461d was done to disable right-click context menus (for issue [#12](../../pywebview/issues/12)).  But currently on MacOS, the context menu can still be activated by ctrl-left-clicking.  I think maybe it used to be disabled in both cases, but the code was changed later, maybe because the original code used a method that's now deprecated?  (webView_contextMenuItemsForElement_defaultMenuItems_ was overridden to return nil, which clears/prevents the menu.)

The current code disables the menu by overriding WKWebView.rightMouseDown_ and does not call the super (unless `_debug['mode']` is true), which would normally activate the menu.

In addition to the fact that ctrl-left-clicking still activates the menu, a second problem I ran into is that circumventing the built-in rightMouseDown_ prevents javascript "mousedown" and "pointerdown" event handlers from seeing right-clicks.

This PR proposes to fix both problems by removing the rightMouseDown_ override, and adding a new override of willOpenMenu_withEvent_ that clears/prevents the menu regardless of what triggered it.

### DIFF
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -276,7 +276,8 @@ class BrowserView:
             super(BrowserView.WebKitHost, self).mouseDown_(event)
 
         def willOpenMenu_withEvent_(self, menu, event):
-            menu.removeAllItems()
+            if not _debug['mode']:
+                menu.removeAllItems()
 
         def performKeyEquivalent_(self, theEvent):
             """
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -275,10 +275,8 @@ class BrowserView:
 
             super(BrowserView.WebKitHost, self).mouseDown_(event)
 
-        def rightMouseDown_(self, event):
-            i = BrowserView.get_instance('webkit', self)
-            if _debug['mode']:
-                super(BrowserView.WebKitHost, self).rightMouseDown_(event)
+        def willOpenMenu_withEvent_(self, menu, event):
+            menu.removeAllItems()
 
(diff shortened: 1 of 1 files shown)


webview/platforms/cocoa.py
```diff
@@ -275,10 +275,8 @@ def mouseDragged_(self, event):
 
             super(BrowserView.WebKitHost, self).mouseDown_(event)
 
-        def rightMouseDown_(self, event):
-            i = BrowserView.get_instance('webkit', self)
-            if _debug['mode']:
-                super(BrowserView.WebKitHost, self).rightMouseDown_(event)
+        def willOpenMenu_withEvent_(self, menu, event):
+            menu.removeAllItems()
```
r0x0r, line 279: Please put debug mode back. Right click menu is visible in debug mode by design.

r0x0r (changes requested):

webview/platforms/cocoa.py
ecpost: Can I put the debug check in the new willOpenMenu_withEvent_ rather than adding back the rightMouseDown_, since that would revert back to blocking javascript events.  In other words (with a "not"), just:

        def willOpenMenu_withEvent_(self, menu, event):
            if not _debug['mode']:
                menu.removeAllItems()

ecpost: ...If that sounds good to you, the latest commit has that.

Issue #996, opened 2022-12-03
Unable to handle and manage special keyboard combinations on MacOS.


jczic:
### Specification
- pywebview version: 3.7.2
- operating system: M1, MacOS Ventura 13.0
- web renderer: Cocoa

### Description  
  
Unable to handle (and/or cancel the propagation of) special keyboard combinations with the MacOS **Cmd/Meta** key like **Meta+Z** or **Meta+Q**.
These events are directly and simply captured by the system but not by the PyWebView UI window.

Example: In javascript "**keydown**", "**keypress**" or "**keyup**" does not catch the combination and the program exits directly or the main window receive signal to close.

(I have integrated the CodeMirror javascript editor and the undo or redo methods are therefore impossible and quit the program).
  
### Practicalities

- [NO] : I am willing to work on this issue myself.

- [NO] : I am prepared to support this issue financially.

ecpost:
I ran into a related (but different) problem with other keys like Home and End, and tracked it down to the `performKeyEquivalent_()` method in cocoa.py.  I also ran into the same problem as you with keys like Meta+C.  Some comments about both types of problems:
-- Our Meta+Z/Meta+C problem is happening because `performKeyEquivalent_()` is intercepting those to translate them into the built-in undo/copy functionality.  The method also returns True to tell the caller that they were already handled, which is why they never make it to the javascript "keydown" event.
-- My Home/End problem is because `performKeyEquivalent_()` only returns False (allowing it to reach the JS event) in certain scenarios.  Specifically, for Meta-key combinations that are NOT intercepted for built-in functionality, and for a hard-coded list of cursor keys (that happens to be missing Home and End).  This is because the method has `return True` at the end, which at first glance looks like a bug.  But I only say "looks like" because this was intentionally added at some point (5f8827537b5ce722960f574209a5631de4e71950), and because of the fact that the list of permitted cursor keys was added to circumvent this same problem for those other keys (ceb647f8bbe7769b031415abfa1e9a4018eb364b).  Although I can't tell from the commit comments if that was thought through all the way or just done as a quick fix.
[...]

PR #999, opened 2022-12-07
Return False from performKeyEquivalent_ in unhandled cases.  Remove cursor workarounds.

ecpost:
Fixes 2nd issue in [#996](../../pywebview/issues/996), allowing any key combinations _not_ handled by `performKeyEquivalent_()` to reach javascript event handlers.


jczic:
👍🏻

PR #1004, opened 2022-12-19
4.0

robb-brown:
Support for WSGI

Encapsulated server in an object so new ones could be added easily. Each window with a local URL or app gets its own server, plus one global server so serverless windows can use the JS API.

Passes tests on a Mac, and runs a non-trivial FastAPI app converted to WSGI with a2wsgi.


webview/window.py
```diff
@@ -62,7 +62,8 @@ class EventContainer:
 class Window:
     def __init__(self, uid, title, url, html, width, height, x, y, resizable, fullscreen,
                  min_size, hidden, frameless, easy_drag, minimized, on_top, confirm_close,
-                 background_color, js_api, text_select, transparent, zoomable, draggable, localization):
+                 background_color, js_api, text_select, transparent, zoomable, draggable, localization,
+                 http_port=None,server=None,serverArgs={}):
```
r0x0r, line 66: commas, server_args

webview/window.py
```diff
@@ -88,9 +89,15 @@ def __init__(self, uid, title, url, html, width, height, x, y, resizable, fullsc
         self.draggable = draggable
         self.localization_override = localization
 
+        # Server config
+        self._http_port=http_port
+        self._server=server
+        self._serverArgs=serverArgs
+
```
r0x0r, line 96: As you moved server parameters to `create_window`, how will it play with `webview.start(http_server=False)`?

webview/__init__.py
```diff
@@ -22,7 +22,7 @@
 import webview.http as http
```
r0x0r, line 22: Please add an example of multi-server http server setup.

r0x0r (changes requested):

r0x0r:
@robb-brown Would you be able to address this PR any time soon? I would like to release 4.0 asap and this would make a nice addition.

robb-brown:
Hi,

I've pushed commit 369a527. To answer your questions:

is_app(url) returns true if the url is a web app. This was a bit of a placeholder, just checking whether the url was not a string, but I've improved it to return true if url is a callable. I think this should cover both WSGI and ASGI web apps, whether they're objects or functionals.

I've added the server and server_args parameters to both create_window and start. create_window passes these along to the window initialization, which stores them. If start has already been called, the window uses them to start its server (if necessary). When start is called, it uses it's parameters to start the global server (if necessary), then each window can start its own server using it's stored object parameters.

Also added examples/multiple_servers.py. This shows how you can start windows before and after calling start. The first window is just plain HTML and so uses the global server. The second is a web app and starts its own server. The third window is created after start is called and is also a web app, so it creates its own server. The third window displays a table of all the windows, their server objects and the IP addresses they're attached to.

robb-brown:
Sure.

PR #1023, opened 2023-01-30
feat: macos supports manual switching of appearance

CahierX:
On macos, you can manually switch the appearance,  setting always light, always dark, or follow the system
such as WeChat:
image

### usage
```python
# theme： light, dark, system
window.set_theme(theme), 
```


r0x0r:
Bump

Issue #1034, opened 2023-02-07
SSL/TLS support over localhost


keredson:
### Specification
- pywebview version:  all
- operating system: all
- web renderer: all

### Description  
  
Any thought to adding SSL to the local server?  Could generate self-signed certs (client and server), then load them into the GTKWebview (or others) for that window only (if possible).  Then the localhost traffic wouldn't be sniffable to anyone w/ the perms, and only that window would be able to connect.
  
### Practicalities

- YES I am willing to work on this issue myself.

- NO I am prepared to support this issue financially.

r0x0r:
This would be a nice improvement. Is it possible to make browser accept self-signed certificates without warnings?

keredson:
I don't know, I'll look into it.  :)

keredson:
Works for the GTK webview at least:
https://github.com/r0x0r/pywebview/compare/master...keredson:pywebview:ssl-proof-of-concept

What do you think about `webview.start(ssl=True)`?

r0x0r:
Sounds good to me

keredson:
Cool, will do a pr, thanks!

r0x0r:
WebView2 allows now ignoring invalid certificates, so I added SSL support to Windows as well. It can be found in the 5.0 branch.

PR #1039, opened 2023-02-09
SSL support for internal server

keredson:
I merged w/ your new threaded adapter, but this doesn't use it in SSL mode (yet).

This *only* supports the GTK backend ATM.  (I'm traveling, don't have access to test anything else ATM.)  Will look into the API for the others later.  I intentionally left it so it will fail (as `add_tls_cert()` won't exist) if you try to use it with another backend. 

This is quite a bit bigger than I anticipated.  I'm not familiar w/ bottle's internals, there may be a better way to wrap the socket w/ bottle mixins.

https://github.com/r0x0r/pywebview/issues/1034

### DIFF
--- a/docs/guide/api.md
+++ b/docs/guide/api.md
@@ -47,7 +47,7 @@ server=http.BottleServer, server_args
 ``` python
 webview.start(func=None, args=None, localization={}, gui=None, debug=False, http_server=False,
               http_port=None, user_agent=None, private_mode=True, storage_path=None, menu=[],
-              server=http.BottleServer, server_args={}):
+              server=http.BottleServer, ssl=False, server_args={}):
 ```
 
 Start a GUI loop and display previously created windows. This function must be called from a main thread.
@@ -64,6 +64,7 @@ Start a GUI loop and display previously created windows. This function must be c
 * `storage_path` - An optional location on hard drive where to store persistant objects. By default `~/.pywebview` is used  on *nix systems and `%APPDATA%\pywebview` on Windows.
 * `menu` - Pass a list of Menu objects to create an application menu. See [this example](/examples/menu.html) for usage details.
 * `server` - A custom WSGI server instance. Defaults to BottleServer.
+* `ssl` - If using the default BottleServer (and for now the GTK backend), will use SSL encryption between the webview and the internal server.
 * `server_args` - Dictionary of arguments to pass through to the server instantiation
 
 ### Examples
--- a/examples/localhost_ssl.py
+++ b/examples/localhost_ssl.py
@@ -1,0 +1,6 @@
+import webview
+
+if __name__ == '__main__':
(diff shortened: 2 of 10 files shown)


keredson:
https://stackoverflow.com/questions/68347874/webview2-server-certificate-custom-validation claims it's not possible w/ WinForms / Chromium

keredson:
bah.  cefpython doesn't support python 3.10:
image

looks like 3.9 is the max:
https://github.com/cztomczak/cefpython/blob/master/src/common/cefpython_public_api.h

keredson:
@r0x0r summary:

- GTK/QT work
- Cocoa works by allowing all certs (no verification, but still useful IMO)
- Webkit.WinForms didn't find any examples, and one claim it's not possible
- CEF didn't work w/ the default python version in win11
- MSHTML is deprecated, so didn't look into
- so all windows renderers throw `NotImplementedError`
- the test failure is the same as in master (timeout - looks like MSHTML is hanging)
- I didn't touch the client cert side of things yet

I'm way over the time-budget I intended for this, I really only care about/need the GTK version, and my macos/windows knowledge is pretty pathetic, so I'm prob. going to leave this as-is for now.  Your call if you want to merge w/o full renderer coverage (and let someone else who cares about windows add to it) or just keep the PR around for someone else to use as a starting point.

Project's great BTW, thanks!

r0x0r:
Good work @keredson 

I found this example for Winforms. Seems like it is possible to catch certificate error and set to ignore it.

https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.servercertificateerrordetected?view=webview2-dotnet-1.0.1518.46

keredson:
@r0x0r no answer, just what i found...

breaking it up:
```
            protection_space = challenge.protectionSpace()
            print('protection_space', protection_space)
            authenticationMethod = protection_space.authenticationMethod
            print('authenticationMethod', authenticationMethod())
            server_trust = protection_space.serverTrust
            print('server_trust', server_trust)
            server_trust_called = server_trust()
            print('server_trust_called', server_trust_called)
            credential = AppKit.NSURLCredential.credentialForTrust_(server_trust_called)
```

accessing each of these properties returns a `native-selector` (i assume a pyobjc thing) which has to be called to get the actual value.

the warning is emitted on `server_trust()`.  it is not on `authenticationMethod()` or any other property (tho that's the only non-primitive that's not null).

NSURLProtectionSpace: https://developer.apple.com/documentation/foundation/nsurlprotectionspace

google only finds https://github.com/ronaldoussoren/pyobjc/issues/493 and https://github.com/ronaldoussoren/pyobjc/issues/298

r0x0r:
I found a solution to Cocoa pyobjcpointer warnings.

Windows seems doable, but it requires a) adding a certificate to the global certificate store b) administrator permissions to do so. Both are rather cumbersome, so let's leave Windows out of this for now.

Issue #1081, opened 2023-03-27
NoneType has no len() and  WebView2 (E_ACCESSDENIED) exceptions


al-eax:
im currently working on a way to add icon support for pywebview. I stumbled over these issues:

### Specification
- pywebview version: current master branch at commit https://github.com/r0x0r/pywebview/commit/9257c25a1e6ae35696b872de77e12d347d33fa04
- operating system: Windows 10
- web renderer:  winforms

### Description  

running the following example i get two exceptions:

```py
import webview
webview.create_window("foo",url="https://google.com")
webview.start()
```

### first exception `NoneType has no len()`
```
Python.Runtime.PythonException: object of type 'NoneType' has no len()
   bei Python.Runtime.PythonException.ThrowLastAsClrException()
   bei Python.Runtime.Dispatcher.TrueDispatch(Object[] args)
   bei Python.Runtime.Dispatcher.Dispatch(Object[] args)
   bei __System_Threading_ThreadStartDispatcher.Invoke()
   bei System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, O
   bei System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object st
   bei System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object st
   bei System.Threading.ThreadHelper.ThreadStart()
```
 
__caused by:__

https://github.com/r0x0r/pywebview/blob/9257c25a1e6ae35696b872de77e12d347d33fa04/webview/platforms/winforms.py#L189

__simple fix:__

```py
if BrowserView.app_menu_list:
     self.set_window_menu(BrowserView.app_menu_list)
```
or set `app_menu_list = []` in https://github.com/r0x0r/pywebview/blob/9257c25a1e6ae35696b872de77e12d347d33fa04/webview/platforms/winforms.py#L135

### second exception `E_ACCESSDENIED`

```
[pywebview] WebView2 initialization failed with exception:
 Zugriff verweigert (Ausnahme von HRESULT: 0x80070005 (E_ACCESSDENIED))
   bei System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
   bei Microsoft.Web.WebView2.Core.CoreWebView2Environment.<CreateAsync>d__3.MoveNext()
[...]

PR #1082, opened 2023-03-27
Fix edgechromium cache

al-eax:
This PR solves two bugs described in #1081.

### first bug

an uninitialized list was checked for its length.

### second bug

`_private_mode` is `True` by default. This causes `cache_dir` to be `None` by default in `winforms.py`
Webview2 crashes if no `UserDataFolder` exists.

To solve this issue, pythons `tempfile.TemporaryDirectory` is used to set `cache_dir` to a temporary directory.

Actually, I wanted to use `TemporaryDirectory` context manager to auto delete the created directory, after window is closed.
Unfortunately, the process still accesses some of the files and blocks the deletion process. A `sleep(1)` solves this issue, but that's a dirty solution.

### DIFF
--- a/webview/platforms/edgechromium.py
+++ b/webview/platforms/edgechromium.py
@@ -49,7 +49,9 @@ class EdgeChrome:
         self.web_view = WebView2()
         props = CoreWebView2CreationProperties()
         props.UserDataFolder = cache_dir
+        props.set_IsInPrivateModeEnabled(_private_mode)
         self.web_view.CreationProperties = props
+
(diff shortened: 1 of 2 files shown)


al-eax:
Im not an test expert when it comes to testing.
Couls some one help me to fix the failed test:
`Build execution time has reached the maximum allowed time for your plan (60 minutes).`

al-eax:
@r0x0r what can I improve to make this PR more reviewable?

PR #1154, opened 2023-06-13
Add missing macOS dependency (fix crash in pywebview 4.1)

simonrob:
df72ac1 added a new import to address a console warning on macOS, but did not add the corresponding requirement `pyobjc-framework-Security`. As a result, pywebview 4.1 crashes on macOS with the following message: `*** Terminating app due to uncaught exception 'OC_PythonException', reason: '<class 'ModuleNotFoundError'>: No module named 'Security''`.

Installing `pyobjc-framework-Security` fixes this issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 pythonnet; sys_platform == "win32"
 pyobjc-core; sys_platform == "darwin"
 pyobjc-framework-Cocoa; sys_platform == "darwin"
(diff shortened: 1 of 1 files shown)


r0x0r:
This fix is already in master. Thanks anyway!

r0x0r:
I made a fix only to pyproject.toml. You are right about requirements.txt

PR #1162, opened 2023-06-27
Support initializing a window in the maximized state.

### LINKED ISSUE #917: How to maximize window on startup?
### Specification
- pywebview version:  3.6.3
- operating system: windows 7 64
- web renderer: edge

### Description  
  
 I want to maximize the window when the app is launched (* I don't want to make it full-screen), but it seems there is no method available to do it

image

 Is this feature still not implemented ?

vsajip:
This should fix #917. Smoke tested on Linux with Qt.


webview/platforms/qt.py
```diff
@@ -790,7 +790,10 @@ def _create():
 
         _main_window_created.set()
 
-        if window.minimized:
+        if window.maximized:
+            browser.showNormal()
```
r0x0r, line 794: Is showNormal needed?

r0x0r (changes requested):

webview/platforms/qt.py
vsajip: Actually, it isn't. I've removed it.

Issue #1178, opened 2023-07-23
Window closing event fired after closing confirmation


p4bl0-:
### Specification
- pywebview version: 4.2.2
- python: 3.11.2
- operating system: Linux Debian 12.0
- web renderer: GTK (Qt is not working and can only display a blank window — note that I'm using KDE)

### Description

The [`events.closing`](https://pywebview.flowrl.com/guide/api.html#events-closing) documentation says : 

> Event fired when pywebview window is about to be closed. **If confirm_quit is set, then this event is fired _before_ the close confirmation is displayed**. If event handler returns False, the close operation will be cancelled.

This does not seem to be the case:

```python
import webview
def bug ():
    print("hey")
window = webview.create_window('events.closing bug', confirm_close=True)
window.events.closing += bug
webview.start()
```

Expected behavior: when closing the window, see `hey` printed on stdout _then_ see the confirmation dialog.

Actual behavior: when closing the window, the confirmation dialog appears and if canceled, nothing happens, if confirmed, the window gets closed and `hey` is printed on stdout.

The behavior is the same with `events.closed`.

Note that if the `bug` function returns `False` however, the behavior does not change using the `closed` event but it prevents the window from being closed using the `closing` event, so at least this works as expected.

However, if `confirm_close` is not set then returning `False` from the closing-event handler does not prevent the window from being closed, and that's probably a bug too.

p4bl0-:
I'd like to add that I found this bug while doing some tests because I wanted to be able to issue a specific confirmation dialog (something like "you have unsaved work, do you wish to save it before you quit?"), but trying to create a confirmation dialog while the window is closing does not work and makes the program hang indefinitely.

I tried create a new window to open the confirmation dialog from it but it still has the same behavior (even without the confirmation dialog, just creating a new window dring the closing event makes the whole program hand indefinitely).

p4bl0-:
> However, if confirm_close is not set then returning False from the closing-event handler does not prevent the window from being closed, and that's probably a bug too.

Fixed this part in #1179 for GTK (I don't know if other GUI lib are affected).

p4bl0-:
#1180 fixes this issue.

p4bl0-:
With #1181 merged I believe this issue can be closed.

PR #1180, opened 2023-07-23
[GTK] closing event working as documented

p4bl0-:
This is what I view as a complete fix for the closing event beahvior under GTK. I did a separate pull request from a smaller fix because I'm not sure this one is wanted: maybe the documentation is wrong and the code was right.

What this does is to actually run the `events.closing` handlers _before_ the closing confirmation (if `confirm_close` is set), as explained in the documentation (see issue #1178).

It also simplifies the GTK code by merging the `close_window` and `on_destroy` functions, in the spirit of how it is done in the Qt code.

Maybe the new merged function should rather be called `on_window_close` to match other event handlers. But that's not really for me to decide :).

-----

As a bonus, it also looks up the value of the `confirm_close` property on the pywebview window at event time rather than initialization time, so that this can be dynamically changed (see the first comment of issue #1178).

This new behavior should be backward compatible with the previous one except if some code exists that changed the value of the `confirm_close` property of their pywebview windows with no effect previously (but why would anyone do that?).

However, this new behavior will differ on other backend as is, as I could only test it with GTK so that's the only one I changed in this PR. But note that changing this behavior for other backends as well should be very straightforward since they all have access to the pywebview window.

### DIFF
--- a/docs/guide/api.md
+++ b/docs/guide/api.md
@@ -382,7 +382,7 @@ Event fired just before pywebview window is closed.
 [Example](/examples/events.html)
 
 ## events.closing
-Event fired when pywebview window is about to be closed. If confirm_quit is set, then this event is fired before the close confirmation is displayed. If event handler returns False, the close operation will be cancelled.
+Event fired when pywebview window is about to be closed. If confirm_close is set, then this event is fired before the close confirmation is displayed. If event handler returns False, the close operation will be cancelled.
 
 [Example](/examples/events.html)
 
--- a/webview/platforms/gtk.py
+++ b/webview/platforms/gtk.py
@@ -195,7 +195,7 @@ class BrowserView:
         should_cancel = self.pywebview_window.events.closing.set()
 
         if should_cancel:
-            return
+            return True
 
         for res in self.js_results.values():
             res['semaphore'].release()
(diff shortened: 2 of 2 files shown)


p4bl0-:
> As a bonus, it also looks up the value of the confirm_close property on the pywebview window at event time rather than initialization time, so that this can be dynamically changed (see the first comment of issue https://github.com/r0x0r/pywebview/issues/1178).
> 
> This new behavior should be backward compatible with the previous one except if some code exists that changed the value of the confirm_close property of their pywebview windows with no effect previously (but why would anyone do that?).
>
> However, this new behavior will differ on other backend as is, as I could only test it with GTK so that's the only one I changed in this PR. But note that changing this behavior for other backends as well should be very straightforward since they all have access to the pywebview window.

This was actually already the behavior in the Cocoa and Windows implementations!

See [here](https://github.com/r0x0r/pywebview/blob/cec4e38e0f3a73512c8cd81bd36f9bdd84782621/webview/platforms/winforms.py#L282) for windows, and [here](https://github.com/r0x0r/pywebview/blob/cec4e38e0f3a73512c8cd81bd36f9bdd84782621/webview/platforms/cocoa.py#L914) and [there](https://github.com/r0x0r/pywebview/blob/cec4e38e0f3a73512c8cd81bd36f9bdd84782621/webview/platforms/cocoa.py#L82) for Cocoa.

p4bl0-:
> What this does is to actually run the events.closing handlers before the closing confirmation (if confirm_close is set), as explained in the documentation (see issue #1178).

As tagged in the title, this PR only fixes the GTK backend.

Please anyone with either a Windows, a Mac, or a Linux with a working Qt backend test #1181 so I can mark it as ready :).

PR #1181, opened 2023-07-24
Update closing event behavior to match GTK behavior in PR #1180

p4bl0-:
This is a sibling PR to #1180, but update Qt, Cocoa, and Windows platforms in addition to GTK. The goal is to make the implementation follow the documentation which states that window closing event handlers are called before the closing confirmation dialog (in the case `confirm_close` is set).

I made it a separate PR because even though the changes are quite straightforward and should not cause any problems, _**I have not tested them**_. I do not have a Windows nor a Mac, and the Qt platform does not work for me (I'm using KDE but strangely, attempting to use Qt with pywebview fails, it only show a blank window, that's why I work using GTK).

So, these have to be tested by someone before they get merged!

Here is a simple test program:

```python
import time
import webview

prevent_closing = True

def on_closing ():
    global prevent_closing
    return not prevent_closing

def test (window):
    global prevent_closing
    print("prevent closing for 5s")
    time.sleep(5)
    prevent_closing = False
    print("can now close, but confirmation required for 5s")
    time.sleep(5)
    window.confirm_close = False
    print("can now close without confirmation")

window = webview.create_window('Test PR #1181', confirm_close=True)
window.events.closing += on_closing
webview.start(test, window)
```

### DIFF
--- a/docs/guide/api.md
+++ b/docs/guide/api.md
@@ -382,7 +382,7 @@ Event fired just before pywebview window is closed.
 [Example](/examples/events.html)
 
 ## events.closing
-Event fired when pywebview window is about to be closed. If confirm_quit is set, then this event is fired before the close confirmation is displayed. If event handler returns False, the close operation will be cancelled.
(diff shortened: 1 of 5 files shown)


r0x0r:
I have tested other platforms, everything works fine. Good work!

PR #1185, opened 2023-07-28
FastApi exemple

xoheveras:
app-server-architecture


r0x0r:
The example cannot be run 
`ERROR:    Error loading ASGI app. Could not import module "app".`

PR #1221, opened 2023-09-21
Update api.md load_url issue

Zirpon:
load_url issue

if url is a local path file, abspath recommended. 
Otherwise, relative path would load fail after freezing the app

### DIFF
--- a/docs/guide/api.md
+++ b/docs/guide/api.md
@@ -306,6 +306,8 @@ window.load_url(url)
 ```
 
 Load a new URL.
+if url is a local path file, abspath recommended. 
+Otherwise, relative path would load fail after freezing the app
 
 [Example](/examples/change_url.html)
 


r0x0r:
This sounds like a bug. Local urls should be resolved correctly. What platform / freezing solution are you using?

Zirpon:
> This sounds like a bug. Local urls should be resolved correctly. What platform / freezing solution are you using?

window python 3.11.0  pipenv Pyinstaller 
details refering my app [app](https://github.com/Zirpon/majSoulRecordPrint)

- [ ] src/menuApi.py
```python
    record_url = '../assets/record.html'
    majSoul_url = 'https://game.maj-soul.com/1/'
    react_url = '../assets/react.html'

    ...

    @staticmethod
    def view_react():
        active_window = webview.active_window()
        active_window.set_title(MenuApi.ReactWindowsTitle)
        active_window.load_url(MenuApi.react_url)
```

r0x0r:
I could not reproduce this. I used modified `examples/change_url.py`

```
import time
import webview


def change_url(window):
    # wait a few seconds before changing url:
    time.sleep(10)

    # change url:
    window.load_url('assets/cookies.html')


if __name__ == '__main__':
    window = webview.create_window('URL Change Example', 'assets/index.html')
    webview.start(change_url, window)
```

Froze with `pyinstaller --onefile  --add-data "./examples/assets/*;assets" .\examples\test.py`. Works without onefile too.

Zirpon:
> I could not reproduce this. I used modified `examples/change_url.py`
> 
> ```
> import time
> import webview
> 
> 
> def change_url(window):
>     # wait a few seconds before changing url:
>     time.sleep(10)
> 
>     # change url:
>     window.load_url('assets/cookies.html')
> 
> 
> if __name__ == '__main__':
>     window = webview.create_window('URL Change Example', 'assets/index.html')
>     webview.start(change_url, window)
> ```
> 
> Froze with `pyinstaller --onefile --add-data "./examples/assets/*;assets" .\examples\test.py`. Works without onefile too.

I think i have figured out what's wrong：
1. when launching pyinstaller freezing app
    - [ ] app will create a folder on the path `C:\Users\Admin\AppData\Local\Temp\_MEI351082`  and set the path to param 'sys._MEIPASS'
    - [ ] refering the pyinstall [doc](https://pyinstaller.org/en/stable/runtime-information.html?highlight=_MEIPASS]
    - [ ] and extract the `--add-data`  files into that path
    - [ ] so freezing app should join the `sys._MEIPASS`  with the `relative path in the code`
 2. the reason your propose example run normally, i think maybe is  because the app launch in the original source path 
  - [ ] if move that app out of the source path maybe can't find the asset
  3. solution：

```python
if getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS'):
  print('running in a PyInstaller bundle')
  import LOADING
  GLOBALS.set_value('DEBUG', False)
  GLOBALS.set_value('BASE_PATH',sys._MEIPASS)
else:
  GLOBALS.set_value('DEBUG', True)
  GLOBALS.set_value('BASE_PATH', os.path.abspath(os.curdir))
```

r0x0r:
Pywebview already takes SYS._MEIPASS into account. The relevant code can be found in util.py

Zirpon:
i see , many thanks :thumbsup:

Issue #1013, opened 2023-01-12
Force Access local file (file://) - Except CORS


Xavlight:
### Specification
- pywebview version:  3.7.2
- operating system: Windows 11
- web renderer: 

### Description  
  
I would like to access the local file (file://) using pywebview
This is possible with the --allow-file-access-from-files parameter when passed as a parameter to msedge or chrome.
To force local access.

I tried to indicate it in args.
webview.start(func=None, args='--allow-file-access-from-files', localization={}, gui=False, debug=False, http_server=False, user_agent=None)
But I think it doesn't have to be done for that.

Simple HTML example:
<video id="video" class="" autoplay="" preload="auto" poster="medias/poster_vide.png" controls="">
	<source id="mp4" type="video/mp4" src="medias/video1_720p.mp4">
	<track id="st" kind="subtitles" src="medias/video1 (En).vtt" label="En" default>
</video>

The access for the MP4 video works, but not for the subtitle (vtt), error CORS (normal).
If I test with a Chrome or edge with the settings to enable local file access, it works perfectly.

Error HTML :
Unsafe attempt to load URL file:///D:/Projets/spice/demo/medias/video1 (Fr).vtt from frame with URL file:///D:/Projets/spice/demo/index.html. 'file:' URLs are treated as unique security origins.

How can I get my webview to be able to access my files locally through my web page.

Thanks

Xav

r0x0r:
You cannot supply browser command line parameters via the args argument. It is for passing parameters to the callback function. 
File access work for some (older) renderers, but is disabled for newer Chromium and such. Any reason why you want to use it? You might want to try 4.0 branch, where it automatically launches a http server for local relative urls (ie with no file:// prefix).  

Introducing an insecure mode for accessing local files and bypassing CORS could be an option if anybody would like to give it a try.

Xavlight:
Thank you for the quick answer.

Actually, I know it's not correct to use local file access (file://).
But it would be to create a media player with pywebviewer, on simple local machines without web server or others (no user, just media players).
This Player would use HTML to play media locally (mp4, images ...) with all the possibilities of html.
But I will still look at version 4.0.

Thanks a lot.

Xav

mphirke:
For accessing local files, you could try the File System access API introduced couple years back. 

https://developer.mozilla.org/en-US/docs/Web/API/File_System_Access_API

I know you can handle images with that, not sure about videos.

PR #1237, opened 2023-10-23
Enable access to local-files when using file:// uris on omitted platforms.

glorpen:
It enables access to `file://` paths when on `file://` loaded page. It does not allow to access `file://` when using remote or built-in http server.
For supported platforms, QT - going by their docs - seems to enable it by default and Cocoa has in enabled explicitly. Seems that only GTK and Edge are omitted.

I'm not sure if it would be better to have it as an option, but then if defaulted to False things could stop working for people on Mac and QT. So enabling it at least should give same behavior everywhere.

Similar (same?) to https://github.com/r0x0r/pywebview/issues/1013.


r0x0r:
This could be configurable via `webview.settings['ALLOW_FILE_URLS']`. webview.settings is going to be introduced in 5.0.
I am merging this into the 5.0 branch

PR #1258, opened 2023-11-15
Replace deprecated Gdk (since 3.22) & WebKit (since 2.40) functions

starnight:
Some Gdk and WebKit functions have become deprecated.  Replace them for compatibility in the future.


starnight:
This test is passed on Ubuntu 22.04.  But, will be failed on Ubuntu 20.04.
So, this should not be merged until pywebview is not maintained for Ubuntu 20.04.

In other words, Ubuntu 20.04 blocks upgrading to GTK 4.

PR #1262, opened 2023-11-18
[GTK] Fix WebKit 4.0 as fallback

starnight:
Some Linux OS versions have not included WekKit 4.1, such as Ubuntu 20.04. Fix the Webkit 4.0 and Soup 2.4 as the fallback.

Fixes commit 34432a3bc8a3 ("[GTK] Fallback for Webkit4.1")


r0x0r:
I made a similar fix to 4.4.1 but due to the typo it still tried to load 4.1. Thanks for the contribution!

PR #1267, opened 2023-11-22
Add the http host parameter

DmitriDanshin:


r0x0r:
What is the use case for this param?

tusharsadhwani:
@r0x0r running the server on `0.0.0.0` for example would let the server be accessible via a different machine on the network. Could be useful for debugging.

But I'm just guessing.

PR #1273, opened 2023-11-26
[appveyor] Separate QT & GTK tests

starnight:
The original tests with Ubuntu2004 image only execute QT part. Separate them with configuration qt & gtk matrix to have both tests.


starnight:
Looks like it gets syntax error.

PR #1275, opened 2023-11-26
[appveyor] Use Ubuntu 22.04 to test GTK platform

starnight:
The original appveyor matrix uses two Ubuntu2004 images to test QT and GTK platforms. But, that might confuse the runner and always test QT platform. So, change GTK platform runner's image to Ubuntu2204.


r0x0r:
I never had a chance to finish creating Appveyour conf. Thanks for the PR!

PR #1294, opened 2024-01-03
Support for nested API definitions

AlecHaring:
This PR introduces nested API definitions. Previously, only methods defined on the base API class instance were available on the frontend. The new structure allows for a more organized, readable, and scalable approach by grouping related functionalities into separate classes.

**Example Usage**:
```python
class MathAPI:
    @classmethod
    def add(cls, a: int, b: int):
        return a + b

    @classmethod
    def subtract(cls, a: int, b: int):
        return a - b

class GreetingAPI:
    @classmethod
    def greet(cls, name: str):
        return f"Hello, {name}!"

class RootAPI:
    math = MathAPI
    greetings = GreetingAPI

    @classmethod
    def ping(cls):
        return "pong"

if __name__ == '__main__':
    window = webview.create_window('JS API example', html=html, js_api=RootAPI)
    webview.start()
```

With this structure, the APIs can be accessed in JavaScript as follows:
- `pywebview.api.ping`
- `pywebview.api.math.add`
- `pywebview.api.math.subtract`
- `pywebview.api.greetings.greet`

**Benefits**:
- **Enhanced Organization**: Group related functionalities into classes.
- **Improved Readability**: Easier to navigate and understand the API structure.
- **Scalability**: More scalable for applications that require a large number of API endpoints.

### DIFF
--- a/webview/js/api.py
+++ b/webview/js/api.py
@@ -5,21 +5,28 @@ window.pywebview = {
     api: {},
 
     _createApi: function(funcList) {
-        for (var i = 0; i < funcList.length; i++) {
-            var funcName = funcList[i].func;
-            var params = funcList[i].params;
+        funcList.forEach(({ func: funcName, params }) => {
+            // Create nested structure and assign function
+            const funcHierarchy = funcName.split('.');
+            const functionName = funcHierarchy.pop();
+            const nestedObject = funcHierarchy.reduce((obj, prop) => {
+                return obj[prop] = obj[prop] || {};
+            }, window.pywebview.api);
 
-            var funcBody =
-                "var __id = (Math.random() + '').substring(2); " +
-                "var promise = new Promise(function(resolve, reject) { " +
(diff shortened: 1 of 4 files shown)


r0x0r:
Could you update `examples/js_api.py` as well?

webview/js/api.py
```diff
@@ -5,21 +5,28 @@
     api: {},
 
     _createApi: function(funcList) {
-        for (var i = 0; i < funcList.length; i++) {
-            var funcName = funcList[i].func;
-            var params = funcList[i].params;
+        funcList.forEach(({ func: funcName, params }) => {
```
r0x0r, line 8: Even though MSHTML is deprecated, it would be nice to preserve minimal compatibility. So please no arrow functions or string literals.

AlecHaring: Understandable. The newest commit fixes this.

AlecHaring:
> Could you update `examples/js_api.py` as well?

Updated :)

PR #1295, opened 2024-01-05
fixed background color not available in webview2

Vanthys:
I noticed that setting the background_color property does not actually change anything on windows except for the "transparent" background setting. There still are a few problems for setting different alpha values, but normal rgb hex values are no problem, and this proves it.


Vanthys:
there would also be the option to use `System.Drawing.ColorTranslator.FromHtml(hex);` but it is actually less ideal, as it _needs_ the '#' before the values

PR #1381, opened 2024-05-14
Add `py.typed` marker

lanzz:
`mypy` is complaining that `webview` doesn't have a `py.typed` marker file, even though the codebase seems to be type annotated already. This PR adds the marker.


lanzz:
Issue: https://github.com/r0x0r/pywebview/pull/1381

PR #1399, opened 2024-05-30
Drag and Drop Lag Issue Resolved

Nexlab-One:


Nexlab-One (commented):
Refactored code to remove lag

PR #1416, opened 2024-06-19
Add support for SPA pages

AzielCF:
## Hello!
I recently discovered this project and I liked it a lot. While researching its features and testing it, I noticed a detail related to the SPAs. 
So before waiting for an update I ventured to make a branch and see what happened. 
Fortunately I was able to solve it with few changes. I hope I can be of help for the project :D/

## Problem:
I noticed that the vue navigation bar was not activated at startup.

image

I also noticed that with a Vue project it did not work, and when reloading the page in a new path it failed, but on the other hand it worked with Flask or Vite.

Captura de pantalla 2024-06-18 205056

## Solution:
With the changes made you can now specify for example a folder "dist/" and it will look for the _**index.html**_ by default.:
`webview.create_window('SPA Support', 'dist/')`
This is necessary because when the SPA is opened for the first time, the navbar does not recognize which route it is on. 

image

On the other hand, for the other routes, if some gives us 404 it will continue reading the main index.html, with this when reloading the route it will continue being valid.
Captura de pantalla 2024-06-18 210142

### _**Note: For the SPA to work correctly you must specify only the folder, do not specify the index.html in the path.**_

### DIFF
--- a/webview/http.py
+++ b/webview/http.py
@@ -38,7 +38,7 @@ WST_co = TypeVar('WST_co', bound=WSGIServer, covariant=True)
 
 logger = logging.getLogger(__name__)
 global_server = None
-
+file_index_default = 'index.html'
 
 def _get_random_port() -> int:
     while True:
@@ -97,7 +97,9 @@ class BottleServer:
         else:
             local_urls = [u for u in urls if is_local_url(u)]
             common_path = (
-                os.path.dirname(os.path.commonpath(local_urls)) if len(local_urls) > 0 else None
+                os.path.commonpath(local_urls) if len(local_urls) > 0 and local_urls[-1].endswith('/') else
+                os.path.dirname(os.path.commonpath(local_urls)) if len(local_urls) > 0 else
+                None
             )
             server.root_path = abspath(common_path) if common_path is not None else None
             app = bottle.Bottle()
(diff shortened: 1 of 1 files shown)


r0x0r:
@AzielCF while this is a useful addition, I do not agree on making this a default setting. A better approach would be to have a fallback url parameter to `webview.create_window` or `webview.start`.
I am not sure what api should look like. There's an existing `server_args` param which is not really used for anything useful. Maybe it could be reused.

AzielCF:
Entiendo su punto, en todo caso tengo una idea que quizás pueda agradarte mas:
Agregar a webview.create_window un argumento spa_mode=true donde un ejemplo sería así:
`webview.create_window('SPA Support', 'dist/index.html', spa_mode=true)`
En código podríamos separar la url para obtener el index.html o cualquier otro archivo que quiera tomar como "principal" el desarrollador para el servidor.
El servidor seguiría iniciando como en el pr actual por el argumento activo.

AzielCF:
I understand your point, in any case I have an idea that you might like better:
Add the parameter in `server_args={'spa':True}` where an example would look like this:

```
webview.create_window('SPA Support', 'dist/index.html')
webview.start(server_args={'spa':True})
```

In code we could separate the url to get the index.html or any other file we want to take as "main" for the server.
The server would still start as in the current pr for the active argument
`htttp://127.0.0.1/`
 and without it it would continue to run as it does by default.
`htttp://127.0.0.0.0.1/index.html`

This way we avoid specifying in the code the index and we make it customizable and finally we give utility to server_args.
If you like the idea I could work on it, I really like this repo :)/

r0x0r:
Let me ponder that. I am not sure it would be better to get rid of server_args or extend it to be useful. Also a `fallback_url: str` could be a more flexible solution in place of `spa: True`

PR #1436, opened 2024-07-15
Add window.events.request_sent and response_received

hustshenl:
1. Fixed EdgeChromium get_current_url BUG
2. Add window.events.request_sent and response_received
3. Add Events example

### DIFF
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -331,9 +331,12 @@ class BrowserView:
             self.setTarget_(self)
             return self
 
+        def setFileDialog_(self, file_dlg):
+            self.file_dlg = file_dlg
+
         def onChange_(self, sender):
             option = sender.indexOfSelectedItem()
-            self.window().setAllowedFileTypes_(self.filter[option][1])
+            self.file_dlg.setAllowedFileTypes_(self.filter[option][1])
 
     class WebKitHost(WebKit.WKWebView):
         def performDragOperation_(self, sender):
@@ -857,6 +860,7 @@ class BrowserView:
                         filter_chooser = BrowserView.FileFilterChooser.alloc().initWithFilter_(
                             file_filter
                         )
+                        filter_chooser.setFileDialog_(open_dlg)
                         open_dlg.setAccessoryView_(filter_chooser)
                         open_dlg.setAccessoryViewDisclosed_(True)
 
@@ -1154,7 +1158,7 @@ def create_file_dialog(dialog_type, directory, allow_multiple, save_filename, fi
     for s in file_types:
         description, extensions = parse_file_type(s)
         file_extensions = [i.lstrip('*.') for i in extensions.split(';') if i != '*.*']
-        file_filter.append([description, file_extensions or None])
+        file_filter.append([description, file_extensions or []])
 
     i = BrowserView.instances.get(uid)
     return i.create_file_dialog(dialog_type, directory, allow_multiple, save_filename, file_filter)


r0x0r:
I have been lazily working on the implementation of request and response headers. The current status can be found in the `headers` branch. I have introduced `models.Request` and `models.Response` classes.  There is a basic EdgeChrome implementation too, but it misses the request/response data found in your PR. Would you mind using the `headers` branch as the base of your PR?
Have you found a way to edit response headers?

hustshenl:
> I have been lazily working on the implementation of request and response headers. The current status can be found in the `headers` branch. I have introduced `models.Request` and `models.Response` classes. There is a basic EdgeChrome implementation too, but it misses the request/response data found in your PR. Would you mind using the `headers` branch as the base of your PR? Have you found a way to edit response headers?

1.OK, I will work on your `headers` branch.
2.I found a way to edit the response data, but there are still some issues that need to be resolved.
You can use Python's httpx to send a request in on_web_resource_request to get the response data. At this point, modify the response headers and content, and then assign the response data to args.Response. The browser will directly use args.Response to display the result.
However, there is a problem: using httpx to send requests in Python causes the browser UI to freeze. It seems that asynchronous requests in C# are needed to avoid this issue, but I don't know how to handle it.
`
    def on_web_resource_request(self, sender, args):
        original_headers = {}
        for header in args.Request.Headers.GetEnumerator():
            original_headers[header.Key] = header.Value

        request = Request(str(args.Request.Uri), args.Request.Method, original_headers)
        self.pywebview_window.events.request_sent.set(request)

        if request.headers == original_headers:
            return

        missing_headers = {k: v for k, v in request.headers.items() if k not in original_headers or original_headers[k] != v}
        extra_headers = {k: v for k, v in original_headers.items() if k not in request.headers}

        for k, v in extra_headers.items():
            args.Request.Headers.SetHeader(k, v)

        for k in missing_headers:
            args.Request.Headers.RemoveHeader(k)
        py_response = httpx.request(method=method,url=url,.....  #  here should run asynchronous requests in C#, I don't kown how 
        res_headers = [f'{key}:{value}' for key, value in py_response.headers]
        bytes_io = io.BytesIO(py_responsecontent)
        byte_array = Array[Byte](bytes_io.getvalue())
        memory_stream = MemoryStream(byte_array)
        response = self.web_view.CoreWebView2.Environment.CreateWebResourceResponse(
	   memory_stream, py_response.status_code', py_response.reason_phrase, '\n'.join(res_headers));
        args.Response = response
`

webview/js/api.py
```diff
@@ -66,6 +64,12 @@
                     // Full file path support for WebView2
                     if (params.event instanceof Event && params.event.type === 'drop' && params.event.dataTransfer.files) {
                         chrome.webview.postMessageWithAdditionalObjects('FilesDropped', params.event.dataTransfer.files);
+                        let event = {};
+                        for (let i in params.event) {
+                            if (i === 'currentTarget') continue;
+                            event[i] = params.event[i];
+                        }
+                        params.event = event
```
r0x0r, line 72: What is purpose of this code?

webview/util.py
```diff
@@ -251,12 +252,24 @@ def get_nested_attribute(obj: object, attr_str: str):
...
+                for item in _dnd_state['paths']:
+                    if item[0] == file['name']:
+                        full_path = item[1]
+                        path = item
+                        break
+                    if urllib.parse.unquote(item[0]) == file['name']:
+                        full_path = urllib.parse.unquote(item[1])
+                        path = item
+                        break
+                if full_path is None:
```
r0x0r, line 268: What problem do these changes address?

r0x0r (changes requested):
Would you mind making a new PR out of the current url fix?
The request and response code needs be to cleaned up and rebased on top of the headers branch.

webview/js/api.py
hustshenl: if this page has many dom, `pywebview._stringify(params)` will take a few seconds. so we remove `currentTarget`

webview/util.py
hustshenl: if `file['name']` is Chinese text, `item[0]` will be URL encoded, so URL decoding is necessary.

hustshenl:
> Would you mind making a new PR out of the current url fix? The request and response code needs be to cleaned up and rebased on top of the headers branch.

i request #1440. it has been closed some days ago. should i make a new one?

webview/js/api.py
r0x0r: is currentTarget the reason for poor performance?
You could probably simplify this by `delete params.event.currentTarget`

webview/util.py
r0x0r: Got it. `urllib.parse.unquote` can be used with the original code then ie 
`path = [item for item in _dnd_state['paths'] if urllib.parse.unquote(item[0]) == file['name']]`
`file['pywebviewFullPath'] = urllib.parse.unquote(full_path)`

r0x0r:
> 2.I found a way to edit the response data, but there are still some issues that need to be resolved. You can use Python's httpx to send a request in on_web_resource_request to get the response data. At this point, modify the response headers and content, and then assign the response data to args.Response. The browser will directly use args.Response to display the result. However, there is a problem: using httpx to send requests in Python causes the browser UI to freeze. It seems that asynchronous requests in C# are needed to avoid this issue, but I don't know how to handle it.

I think it will be more straightforward to send a request using .net's HttpClient. Response event should be also raised in the request handler after you make a request instead of in `on_web_response_received`.

r0x0r:
> i request #1440. it has been closed some days ago. should i make a new one?

Please make a new PR for dnd and get current url bug fixes with master as the target. Remove unrelated bug fixes from this PR and set this PR's target to `headers`. Thanks.

webview/js/api.py
hustshenl: yes, `currentTarget` is  the reason for poor performance.
I have tried `delete params.event.currentTarget` and `params.event.currentTarget=null`, they do not work.

hustshenl:
> Would you mind making a new PR out of the current url fix? The request and response code needs be to cleaned up and rebased on top of the headers branch.

I have made a new PR #1475

hustshenl:
> > i request #1440. it has been closed some days ago. should i make a new one?
> 
> Please make a new PR for dnd and get current url bug fixes with master as the target. Remove unrelated bug fixes from this PR and set this PR's target to `headers`. Thanks.

I have made a new PR #1476

Issue #1037, opened 2023-02-09
"Access id Denied" webview.create_window().start(); How to configure WebView Default directories. 


CtheCondor:
### Specification
- Latest release from pip Effective 2/9/23
- Windows 10
- web renderer: Unknown?

### Description  
  
  I have confirmed WebView2 is installed on my machine. I do not have ownership of the "C:\" Drive on my machine due to corporate polices. Is there direction you can provide within myWebView to establish a different default folder? I have read https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/user-data-folder?tabs=win32 but I have been unable to translate that into actions.
  
### Practicalities

- [No] I am willing to work on this issue myself. # Don't think its an issue just a configuration/help I need. 

- [NO] I am prepared to support this issue financially.
![image](https://user-images.githubusercontent.com/70120966/217888740-92309c57-d484-4c19-8b3e-0f14e11d1252.png)

CtheCondor:
Hmm in my simple example documented above; by adding the storage_path variable worked however when I tried to add those arguments to the start method of an another pip package that uses pywebview. So I believe this closes this comment for pywebview directly though now I am gonna keep digging into why I am geting issues launching a webview window.

https://github.com/rzel/twpy/blob/master/twpym.py is the project in question that I am struggling to use the integrated webview functionality if interested.

PR #1456, opened 2024-08-23
fix: Set WebView2 user data path

fwRelik:
### Problem:

After building a project using PyInstaller into a single executable file, during the application's runtime, a folder named "exe.WebView2" was created next to the executable file. Due to the location of the executable, the application lacked the necessary permissions to modify its folder, leading to runtime issues.

Attempts to resolve this issue by following the suggestions mentioned in [this discussion](https://github.com/r0x0r/pywebview/issues/1037#issuecomment-1424732083) did not yield the desired results.

### Solution:

Upon consulting the WebView2 documentation, specifically the "Manage user data folders" section [(source)](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/user-data-folder?tabs=dotnet), I discovered that it's possible to specify a custom path for the creation of the user data folder.

I modified the setup_webview2_environment method in the edgechromium.py file to include the `userDataFolder` parameter when calling `CoreWebView2Environment.CreateAsync`. By setting the path to `_settings['storage_path']`, I was able to direct the WebView2 data folder to a location with the appropriate permissions, fully addressing the issue.

This change ensures that the WebView2 data folder is created in a specified.

### DIFF
--- a/webview/platforms/edgechromium.py
+++ b/webview/platforms/edgechromium.py
@@ -77,7 +77,9 @@ class EdgeChrome:
         def _callback(task):
             self.web_view.EnsureCoreWebView2Async(task.Result)
 
-        environment = CoreWebView2Environment.CreateAsync()
+        environment = CoreWebView2Environment.CreateAsync(
+            userDataFolder=_settings['storage_path']
+        )
         environment.ContinueWith(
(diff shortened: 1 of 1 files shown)


gmsnk:
It’s working great.

PR #1475, opened 2024-09-20
fixed some issues

hustshenl:
1. Fix performance issues with the drag-and-drop (dnd) event.
2. Fix the exception caused by `args.get_AdditionalObjects()` being None in certain cases.
3. Fix the Edge URL bug.
4. Fix the bug where the full path cannot be retrieved when dragging a file with Chinese or other Unicode characters.

1. 修复dnd事件性能问题，
2. 修复某些情况下`args.get_AdditionalObjects()`为None导致的异常 
3.修复Edge URL BUG
4.修复拖动文件为中文等Unicode字符时，无法获取完整路径的BUG


r0x0r:
Thanks.
I have removed the drop event slow performance fix for now. While the issue should be definitely fixed, the fix should apply to all the platforms and maybe it can be done in other way than removing currentTarget. I have created an issue for it here https://github.com/r0x0r/pywebview/issues/1485

PR #1478, opened 2024-09-27
Replaced internal URL with permalink

Koladweep:
For some reasons - maybe Github's internal CDN design, the file "pretty" URL isn't enough to access the resource in GitHub when used from external sources. Permalink surely does.  The original link was working from within GitHub but not when it was clicked from PyWebView documentation.


Koladweep:
Visual proof of work
image
image

PR #1479, opened 2024-09-27
Update edgechromium.py

lixuelong135:
I would like to add a setting 'OPEN.ETERNAL_INKSININ_NEW-WINDOW' I need this feature, so I am trying to do it in the hope that it will be helpful for the project


r0x0r:
While this idea is worth doing, please provide a complete PR. I am not going to accept this unless the implementation is complete for all the platforms.

PR #1483, opened 2024-09-30
Handle UNIX paths in drag-and-drop events on GTK

coffeejunk:
The GTK implementation of drag-and-drop events under Linux (tested on Fedora 40, GNOME Shell 46.5) sends a regular UNIX file path (e.g., /home/user/file.txt). However, the current pywebview implementation expects a file:// URI and discards anything else. This leads to pywebviewFullPath being empty despite valid file paths being provided in the drop event.

This commit updates the logic to handle both file:// URIs and standard UNIX paths, ensuring proper drag-and-drop functionality on GTK-based platforms.

https://github.com/user-attachments/assets/1a563870-daa1-46b4-8153-4a698313694d

### DIFF
--- a/webview/platforms/gtk.py
+++ b/webview/platforms/gtk.py
@@ -244,13 +244,18 @@ class BrowserView:
         return False
 
     def on_drag_data(self, widget, drag_context, x, y, data, info, time):
-        if _dnd_state['num_listeners'] > 0 and data.get_text():
-            files = [
-                (os.path.basename(value), value.replace('file://', ''))
-                for value
-                in data.get_text().split('\n')
-                if value.startswith('file://')
-            ]
+        text = data.get_text()
+        if text and _dnd_state['num_listeners']:
+            if text.startswith('/'):
+                # Direct path case (e.g., `/home/user/file.txt`)
+                files = [(os.path.basename(text), text)]
+            else:
+                # 'file://' case (e.g., `file:///home/user/file.txt`)
+                files = [
+                    (os.path.basename(value), value.replace('file://', ''))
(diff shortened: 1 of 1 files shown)


webview/platforms/gtk.py
```diff
@@ -244,13 +244,18 @@ def close_window(self, *data):
...
-        if _dnd_state['num_listeners'] > 0 and data.get_text():
-            files = [
-                (os.path.basename(value), value.replace('file://', ''))
-                for value
-                in data.get_text().split('\n')
-                if value.startswith('file://')
-            ]
+        text = data.get_text()
+        if text and _dnd_state['num_listeners']:
+            if text.startswith('/'):
```
r0x0r, line 249: This check won't do, as it will treat text starting with a slash as a file name. I cannot reproduce this on Ubuntu, but at very least it should maybe check that strings starting with a slash is a valid path name.

webview/platforms/gtk.py
```diff
@@ -244,13 +244,18 @@ def close_window(self, *data):
...
+                # Direct path case (e.g., `/home/user/file.txt`)
+                files = [(os.path.basename(text), text)]
```
r0x0r, line 251: Shouldn't you split you by '\n' first?

webview/platforms/gtk.py
coffeejunk: Makes sense! I've updated the PR.

It's interesting that Fedora and Ubuntu behave differently here. :thinking:

coffeejunk:
Running the DnD example on Fedora:

PR #1490, opened 2024-10-06
Add the feature that having an ability to daemonize the invoked func …

xiaofengsoft:
Add the feature that having an ability to daemonize the invoked func while starting the window

### DIFF
--- a/webview/__init__.py
+++ b/webview/__init__.py
@@ -95,6 +95,7 @@ menus: list[Menu] = []
 
 def start(
     func: Callable[..., None] | None = None,
+    is_func_daemon: bool = False,
     args: Iterable[Any] | None = None,
     localization: dict[str, str] = {},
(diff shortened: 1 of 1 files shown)


r0x0r:
What's the use case for this? 
You can always start a callback manually in whatever way you want.

xiaofengsoft:
Yeah, you re right, but this way may be more expedient. For example, when you need to close the called thread at the same time as closing the window.In this way, It just need you to pass a param.

xiaofengsoft:
> What's the use case for this? 
> You can always start a callback manually in whatever way you want.

Yeah, you re right, but this way may be more expedient. For example, when you need to close the called thread at the same time as closing the window.In this way, It just need you to pass a param.

r0x0r:
Thanks for the suggestion, but I am not going to accept this at this point.

PR #1502, opened 2024-10-18
Allow configuring Chromium remote debugging port

Lugribossk:
Allow configuring the Chromium remote debugging port via `webview.settings`. This makes it possible write tests for the application using Playwright via this guide: https://playwright.dev/docs/webview2


r0x0r:
Have you researched if it is possible to add  remote debugging  for other platforms?

Lugribossk:
It doesn't look like it is possible for now. Playwright does support Webkit but it uses it's [own patched version of it](https://playwright.dev/docs/browsers#webkit) so it will not work with the plain version, which I assume is what gets used on Mac. I think there are some plans to align the Devtools protocol (which is how Playwright communicates with the browser) between Chrome and Webkit, so it might work in the future when this happens.

PR #1511, opened 2024-10-26
Add Chinese translation for documents

moyanj:


moyanj:
I am also think who will keep the Chinese translation up-to-date when the main translator is updated. I can only undertake some translation work in the future.

r0x0r:
A modern solution could be to automatically generate AI assisted translations using a Github action when docs branch is updated

PR #1543, opened 2024-12-10
Implemented Adding Cookies Functionality

elmoiv:
Added the ability to add custom cookies to `EdgeChrome` on NT platforms. This can help in scenarios where the user wants to inject cookies from `requests` session into the current webview.

Example:
```python
window = webview.create_window('Hello world', 'https://example.com')
...
response = requests.post(login_url, data=payload)
...
for cookie in response.cookies:
    window.add_cookie(
        cookie.name,
        cookie.value,
        cookie.path,
        cookie.domain,
    )
```

### DIFF
--- a/tests/test_cookies.py
+++ b/tests/test_cookies.py
@@ -14,6 +14,9 @@ def window():
 def test_get_cookies(window):
     run_test(webview, window, get_cookies_test)
 
+def test_add_cookie(window):
+    run_test(webview, window, add_cookie_test)
+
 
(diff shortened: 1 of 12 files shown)


r0x0r:
What's wrong with `window.evaluate_js('document.cookie = 'newCookie=123'')`?

elmoiv:
Sometimes the easiest solution is in front of your eyes. But you always tend to overwork yourself. Thanks for clarifying a better and more safe way to do so.

PR #1552, opened 2024-12-25
Add support for the download attribute on links for Cocoa/MacOS platform

maddyaby:
The download attribute wasn't forcing a download on mac, this makes it do that, and respects the suggested filename!

Also add a test but skip it since it needs to be run manually and have user confirmation. Test fails before change and passes after.

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -1,0 +1,70 @@
+import pytest
+
+import webview
+
+from .util import run_test, assert_js
+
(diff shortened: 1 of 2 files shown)


maddyaby:
I didn't run any other tests besides the added one - is there a test suite I should run locally or is that all run through CI?

r0x0r:
Thanks!

PR #1556, opened 2024-12-29
Add CEFPython3 to the requirements on Win32

HGStyle:
I had to install CEFPython3 to make the CEF web engine work with pywebview, so I believe it is a missing depency. Correct me if I'm wrong.


r0x0r:
Cefpython3 is an optional dependency and can be installed via `pip install pywebview[cef] `

PR #1567, opened 2025-01-06
[Cocoa] Add handler for javascript prompt/input.

maddyaby:
Cocoa was missing handler for calling prompt("...")

This adds the popup and input handling. It also adds the Edit menu item to handle cut/copy/paste for subviews and modals in addition to the main view


r0x0r:
Prompt works fine, but there is no Edit menu item.

maddyaby:
Like you tried it and you now get the prompt correctly but don't see Edit
up in the mac header bar (next to View)?

I'll double check I pushed everything correctly, and take a screenshot

maddyaby:
rebased and updated

PR #1569, opened 2025-01-08
[Cocoa] Fix location that new subviews get added

### LINKED ISSUE #1568: Native Color Picker on MacOS is drawn in the wrong place
### Specification
- pywebview version: latest/master
- operating system: MacOS
- web renderer: WebKit

### Description

The y-coordinate is inverted when WebKit renders the native color picker menu over the view. This doesn't happen on Safari or Chrome, only in the Webview. 

There's an isFlipped property for the view that might(?) fix it: https://developer.apple.com/documentation/appkit/nsview/isflipped
But I can't figure out how to set that, or if there's any way to intercept the native call to flip the coordinates when it draws.

I can see the mouseDown_ in WebKitHost is firing with the reversed y-coordinate system from what the web sees and those events must be flipped somewhere since the web code sees the correct positions... but the reverse direction when the color picker requests a draw is wrong.

Example:

```
<html>
  <body>
    <input  type="color" value="#e66465" style="position:absolute;" />
    <input  type="color" value="#e66465" style="position:absolute;left:100px;top:100px;"/>
    <input  type="color" value="#e66465" style="position:absolute;top:100px;" />
    <input  type="color" value="#e66465" style="position:absolute;left:100px;" />
  </body>
</html>
```

### Practicalities

- YES I am willing to work on this issue myself.

maddyaby:
[Cocoa] Fix location that new subviews get added

Y axis is flipped for the web coordinate system vs. the window, so events originating from the webview will have flipped coordinates

Resolves #1568

### DIFF
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -363,6 +363,16 @@ class BrowserView:
 
             return super(BrowserView.WebKitHost, self).performDragOperation_(sender)
 
+        def addSubview_(self, event):
+            # translate the event from the web coordinate space into the window coordinate space
+            # basically flip the y axis since the WKWebView/WebKitHost implements a flipped view
+            # (but seems to miss this flip back when adding a subview)
+            event_frame = event.frame()
+            flipped_y = self.frame().size.height - (event_frame.origin.y + event_frame.size.height)
+            event.setFrameOrigin_(AppKit.NSPoint(event_frame.origin.x, flipped_y))
+
+            super(BrowserView.WebKitHost, self).addSubview_(event)
+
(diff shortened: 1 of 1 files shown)


maddyaby:
So this flips the coordinates for any call to addSubview... not sure if we want to restrict it to only ones that originate from the webview (but I don't know how to do that :( ). 

Do we expect any other originators of addSubview beyond the first call that adds the WKFlippedView container for the webview, and calls that originate from the webview itself?

r0x0r:
Good job on fixing this.
I know nothing of subviews, so sorry I don't know an answer to your question

PR #1575, opened 2025-01-11
The window title bar background color adapts to the system theme

godcop:
Currently, the default window title bar of the Windows platform is fixed to white. I added a function to determine whether the system theme is dark or light, so that the title bar can adapt to the system theme.


webview/platforms/winforms.py
```diff
@@ -272,6 +275,21 @@ def __init__(self, window, cache_dir):
         def __str__(self):
             return f'<System.Windows.Forms object with {self.Handle} handle>'
 
+        def get_system_theme(self):
```
r0x0r, line 278: To be more explicit this function could be named `is_dark_theme` and return True or False

godcop: yes, that's right

Issue #1598, opened 2025-02-14
Android fullscreen support


michelle-avery:
I noticed the documentation for Android says that multiple windows aren't supported, but it says nothing about lack of support for fullscreen, while the actual code indicates that fullscreen support actually isn't supported on Android. Is this not feasible due to platform constraints, or is it just something that hasn't been implemented yet?

r0x0r:
I have never thought about fullscreen Android to be honest. It is feasible and should be implemented.

michelle-avery:
Awesome. Is that  something you plan on working on in the near future? If not, I  can take a swing at it. I have a cross-platform project that uses pywebview, and the  Android users would like full-screen capability. I started taking a look at what it would take to implement, but I wanted to make sure it wasn't known to be impossible, or undesirable, before I spent too much time on it, given Android development isn't really my area of expertise.

r0x0r:
Feel free to work on this. The implementation should be no too hard.

PR #1602, opened 2025-03-04
Add fullscreen support to android

michelle-avery:
This PR enables fullscreen support on Android as discussed in #1598


r0x0r:
Great work. Tested to work correctly

PR #1607, opened 2025-03-12
Modify jsapi to use callback instead of setInterval.

qaqFei:


r0x0r:
Care to explain the motivation behind this change and why this is needed.

qaqFei:
If there are many times that are time-consuming and jsapi calls on different threads are time-consuming, setInterval can cause efficiency issues.

jameskeane:
This seems like a nice improvement, good work!

r0x0r:
Sorry for the delay, had to test and make sure everything works. I replaced arrow function with a normal (this is due MSHTML compatibility), but otherwise great work.

Issue #867, opened 2022-03-05
async support for js_api ?


manatlan:
### Specification
- pywebview version:  latest
- operating system: ubuntu latest
- web renderer: html+js

### Description  

Seems that "js_api" support only sync method.
Do you plan to support async method in your loop ?

I would like to do something like that : 
```
        class Api:
            async def interact(self):
                return await mymethod()

        window = webview.create_window("title", html=html, js_api=Api())
```

r0x0r:
There has been an async discussion here  https://github.com/r0x0r/pywebview/issues/801

Async support could be added, but to be brutally honest my async game is not very strong. 
I am open to suggestions how this should be implemented.

manatlan:
In fact, I'm working on [THAG](https://github.com/manatlan/thag)
( Which is a gui toolkit for building gui toolkits )

Here is a demo, with pywebview backend

    $ pip3 install thag pywebview
    $ wget https://raw.githubusercontent.com/manatlan/thag/main/examples/demo.py
    $ python3 demo.py

It works well here, because everything is sync ...
But when "async" call, it can't work with the pywebview backend ;-(

I will have a look in the pywebview code, to try to make it async compliant in a main loop

PR #1609, opened 2025-03-14
Support asyncio in exposed python methods

### LINKED ISSUE #801: Running async calls on evaluate_js or add api called eval_async_js.
### Specification
pywebview version: 3.7
operating system: Window
web renderer: edgechromium, edgehtml, cef,

### Description  
  
Is there a way to run async calls and js promises on evaluate js? fetch and axios returns none or empty {}.  I know that I can just use python request and connect  it with js api, but running async js and promises (not only about fetch and axios) in scrape is important in my case.  It seems evaluate js can only run synchronous calls.  Do it need to communicate to js api to do this like using timeout? Hope you add another api just like on selenium (execute_async_script)  that can run promises async api. It would help scraping alot and give more option on evaluating js on windows.

jameskeane:
This commit adds support for coroutine methods 'async/await'. It does this by simply checking if an exposed method returns a coroutine and if so it awaits it.

This makes it fully backwards compatible, and is purely an enhancement.

Includes a simple test demonstrating how to use it. 

Fixes #801 #867

### DIFF
--- a/webview/__init__.py
+++ b/webview/__init__.py
@@ -18,6 +18,7 @@ import tempfile
 import threading
 from collections.abc import Iterable, Mapping
 from typing import Any, Callable
+from asyncio import AbstractEventLoop
 from uuid import uuid4
 
 from proxy_tools import module_property
@@ -255,6 +256,7 @@ def create_window(
     server: type[http.ServerType] = http.BottleServer,
     http_port: int | None = None,
     server_args: http.ServerArgs = {},
+    event_loop: AbstractEventLoop = None
 ) -> Window:
     """
     Create a web view window using a native GUI. The execution blocks after this function is invoked, so other
@@ -282,6 +284,7 @@ def create_window(
     :param transparent: Don't draw window background.
     :param server: Server class. Defaults to BottleServer
     :param server_args: Dictionary of arguments to pass through to the server instantiation
+    :param event_loop: Optional event loop to support exposing async functions
     :return: window object.
     """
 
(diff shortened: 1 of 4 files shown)


webview/util.py
```diff
@@ -239,6 +240,10 @@ def js_bridge_call(window: Window, func_name: str, param: Any, value_id: str) ->
     def _call():
         try:
             result = func(*func_params)
+            if window.event_loop and asyncio.iscoroutine(result):
```
jameskeane, line 243: We can probably throw an exception here. If the result is a coroutine, but the user hasn't configured the `event_loop` then we could throw an exception explaining they need to configure it.

r0x0r: What benefit does an sync event handler bring here? Since a separate thread is launched for the handler anyway.

jameskeane: The benefit is simply supporting async functions. There are plenty of reasons people might want to expose an existing async API to the bridge calls.

See the test file for usage.

r0x0r: Alright fair enough.

webview/__init__.py
```diff
@@ -255,6 +256,7 @@ def create_window(
     server: type[http.ServerType] = http.BottleServer,
     http_port: int | None = None,
     server_args: http.ServerArgs = {},
+    event_loop: AbstractEventLoop = None
```
r0x0r, line 259: I suggest moving event_loop as a parameter of the `start` function instead. Having a window-dedicated event loop  is a bit overkill imo.

r0x0r (changes requested):
Let's move event_loop to the start function

luojiaaoo:
help！！！
pywebview is a great library, but it's a real pain that it doesn't support asyncio properly. In 2026, async should be a first-class citizen!

PR #1612, opened 2025-03-17
Add new WinUI3 backend.

dlech:
I was looking at using pywebview, but found that it doesn't work on my new ARM64 Windows laptop because Python.NET doesn't support ARM64 yet. I happen to maintain pywinrt which is a Python projection of the Windows SDK and Windows App SDK (WinUI3), so I though I would give it a try to use that to implement it.

The only downside I see is that this requires installing the Windows App Runtime in order to work. So the first time anyone runs script that uses this and they don't have it install already, they will see a dialog box instructing them to install the runtime. For people that package pywebview in an app with an installer can have the installer include the runtime though, so in that case wouldn't really be a problem.

I think I've managed to implement about 80-90% of the features that the .NET backend has. And it is just a matter of someone investing some time to take it the rest of the way. Not sure I will be able to do this any time soon, so happy to let someone else take over if there is interest in this. Enjoy!


r0x0r:
Hey, this is exciting news. I haven't worked with WinUI3, but support for a modern UI framework is a welcomed addition.

Could you elaborate what features are missing from your PR?

dlech:
Unfortunately, I didn't keep notes so I don't have an exact list of what is done and what is left to do.

Since this uses the same WebView2 as the .NET backend, I think most of the JS and cookies type stuff is working since I was able to just copy that code for the most part. Most of the basic window operations like minimize/maximize/fullscreen seem to be working. I don't think I did much to test focus or transparent/frameless windows or drag and drop yet. The file dialogs in WinUI3 aren't great (limited features), so I replaced a few with the classing Win32 calls, but there are a few I haven't converted yet.

So really, what it needs mostly is for someone who is actually familiar with the expected behavior as implemented in the other backends to try all of the examples to say if it looks like it is working correctly or not.

r0x0r:
I tried this, but could not get dependencies installed. I had to install https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/downloads, but other than that I am clueless.
What is needed in order to get it work?

r0x0r:
My bad, I missed installation of Python dependencies via pyproject.yaml. I got `simple_browser.py` working on one machine, but it fails with `[WinError -2147024846] The request is not supported.` on macOS via a virtual machine (Python 3.11 / ARM64). Have you encountered this error?

dlech:
What is the call stack for that error?

What VM do you use on Mac (VMWare, UTM)?

r0x0r:
Parallels 19 on Macbook M1 Pro.
Windows 11 PRO 26100.3476
Tested with a freshly created virtual env.

The stack trace is 
```
  File "\\mac\Home\Code\pywebview\examples\simple_browser.py", line 8, in <module>
    webview.start(gui="winui3")
  File "\\mac\Home\Code\pywebview\webview\__init__.py", line 176, in start
    guilib = initialize(gui)
             ^^^^^^^^^^^^^^^
  File "\\mac\Home\Code\pywebview\webview\guilib.py", line 155, in initialize
    guilib.setup_app()
  File "\\mac\Home\Code\pywebview\webview\platforms\winui3.py", line 930, in setup_app
    initialize(options=InitializeOptions.ON_NO_MATCH_SHOW_UI)
  File "C:\Users\roman\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\winui3\microsoft\windows\applicationmodel\dynamicdependency\bootstrap\__init__.py", line 67, in initialize
    return bootstrap.initialize(ver_maj_min, tag, min_ver_quad, options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [WinError -2147024846] The request is not supported.```

**Comment by dlech** [comment]
No, haven't seen that. But I think it could be caused if you are using the Python runtime installed from the Microsoft Store. I will check it out.

**Comment by dlech** [comment]
OK I confirmed that the problem is that the Microsoft Store version of Python doesn't work because it is a "packaged" app.

There is a `ON_PACKAGE_IDENTITY_NOOP` option flag we can add to `initialize(options=InitializeOptions.ON_NO_MATCH_SHOW_UI)` to avoid the error here, but it doesn't really help because it will just crash later when trying to call a winui3 API because the packaged Python didn't include this in it's package manifest.

```console
  File "C:\Users\extra\work\pywebview\webview\platforms\winui3.py", line 985, in create_window
    Application.start(init_app)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^
AttributeError: method overload with 1 arg(s) is not available in this version of Windows
```

Basically, the same issue as mentioned here: https://stackoverflow.com/questions/76301950/store-app-cant-call-mddbootstrapinitialize

Since we don't control the Python Microsoft Store package, I'm not sure what we could possibly do about this. 😞

dlech:
I managed to find a workaround of sorts, but unfortunately, it only works on Windows 11.

With Windows 10 support ending in October of this year, maybe that isn't so bad? I know some people will be running Windows 10 forever but at least they can work around it by using a Python runtime that was not installed from the Microsoft Store.

webview/platforms/win32.py
```diff
@@ -0,0 +1,220 @@
...
+        path = PWSTR()
+        hr = item.GetDisplayName(SIGDN_FILESYSPATH, path)
+        if FAILED(hr):
+            raise WinError(hr)
+        stack.callback(CoTaskMemFree, path)
+
+        return path.value
+
+
+def win11_bootstrap() -> None:
```
dlech, line 176: This needs a bit more work to return a cleanup function that calls `RemovePackageDependency` and `DeletePackageDependency`.

https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/framework-packages/use-the-dynamic-dependency-api

r0x0r:
I got it up and running by switching to a manually installed Python. I will do more testing later, but here are some issues I have discovered already
- Webview window cannot be scrolled with a trackpad.
- Quite many tests crash. Not sure if it is related to broken functionality or whether that tests are run sequentially.

webview/platforms/winui3.py
```diff
@@ -0,0 +1,1337 @@
...
+    class BrowserForm:
+        def __init__(self, window: _Window, cache_dir: str):
+            self._is_active = False
+            self.uid = window.uid
+            self.pywebview_window = window
+            self.window = Window()
+            self.handle = get_window_from_window_id(self.window.app_window.id)
+            self.real_url = None
+
+            self.pywebview_window.native = self
```
r0x0r, line 521: native should point to a native window (form) instance, ie to `self.window`
Additionally  self.pywebview_window.native.webview should point to a webview object

dlech: fixed

dlech:
> * Webview window cannot be scrolled with a trackpad.

Hmm... I noticed that too. Not sure where to start looking to fix that one.



> Quite many tests crash. Not sure if it is related to broken functionality or whether that tests are run sequentially.

Seems to be related to calling the winui3 bootstrap initialize function multiple times in the same process. I will have to do some more debugging to see what's up.

dlech:
After some more debugging it looks like the crash is happening when we run `Application.start()` for the second time in the same process. I wouldn't be surprised if Microsoft never tested this since Visual Studio puts this in generated code that users never edit. And with the way application lifecycles work, I expect that this is not a bug, it's a feature.

I found a few issues that talk about how the event loop works:
https://github.com/microsoft/microsoft-ui-xaml/issues/4160
https://github.com/microsoft/microsoft-ui-xaml/discussions/9906

To work around this, we will probably have to use Xaml islands in a plain Win32 app instead of using the WinUI application framework. That will take me some time to figure out since I've never done that before.

r0x0r:
@dlech Any progress on this one? At least the scroll issue is a show-stopper, which definitely has to be addressed.

dlech:
No, I haven't had any extra time to work on this for a while. I'll probably come back to it eventually, but I'm not sure when.

r0x0r:
@dlech FYI I am working on this right now. Current progress can be found in the `winui3` branch

PR #1632, opened 2025-05-03
[Cocoa] Add `no_default_menus` start param for omitting default menus

mikeylemmon:
There are cases where the default View and Edit menus don't make sense or aren't desirable. This PR adds an option — `no_default_menus` — to indicate that they shouldn't be created. Example:

    webview.start(menu=menu_items, no_default_menus=True)


mikeylemmon:
@r0x0r let me know if you'd prefer separate params for the View and Edit menus individually, or some other approach entirely — happy to adjust my changes to whatever approach you think is best!

r0x0r:
A better place to control this would be `webview.settings['SHOW_DEFAULT_MENUS']` (True by default)

mikeylemmon:
Updated now to use `webview.settings['SHOW_DEFAULT_MENUS']` and it does seem a lot better, thanks for the feedback!

PR #1661, opened 2025-06-22
[Cocoa] Implement `OPEN_DEVTOOLS_IN_DEBUG`

chrissantamaria:
👋 noticed that `OPEN_DEVTOOLS_IN_DEBUG` wasn't taking effect on Cocoa unlike other GUIs like edgechromium and qt. Took a stab at implementing it, albeit with some private APIs which may or may not work - didn't see any public options available. This worked for me on macOS 15.4.1.

I'm a bit out of my depth here, but some AI assistance nudged that there could be some other possible fallback implementations as well:
```py
try:
    import objc
    _WKInspector = objc.lookUpClass('_WKInspector')
    if _WKInspector:
        inspector = _WKInspector.alloc()
        if hasattr(inspector, 'show'):
            inspector.show()
            return True
except:
    pass

try:
    if hasattr(webview, 'performSelector_'):
        selector_name = objc.selector(b'_showInspector', signature=b'v@:')
        webview.performSelector_(selector_name)
        return True
except:
    pass
```

I left these out given that I haven't tested them, but happy to add back if that's preferable.

Merging this into `6.0` since it's the branch I'm personally working on. I don't see any reason this couldn't be backported to `master` if you're planning another `5.x` release.

### DIFF
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -384,6 +384,9 @@ class BrowserView:
 
                 inject_pywebview('cocoa', i.js_bridge.window)
 
+                if _state['debug'] and webview_settings['OPEN_DEVTOOLS_IN_DEBUG']:
+                    BrowserView._open_web_inspector(webview)
+
         # Handle JavaScript window.print()
         def userContentController_didReceiveScriptMessage_(self, controller, message):
             if message.body() == 'print':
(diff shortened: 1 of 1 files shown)


webview/platforms/cocoa.py
```diff
@@ -635,7 +638,7 @@ def dummy_completion_handler():
 
         config.preferences().setValue_forKey_(webview_settings['ALLOW_FILE_URLS'], 'allowFileAccessFromFileURLs')
 
-        if _state['debug'] and webview_settings['OPEN_DEVTOOLS_IN_DEBUG']:
+        if _state['debug']:
```
chrissantamaria, line 641: Drive-by change, split into a separate commit if we want to drop: I don't think we should be checking `OPEN_DEVTOOLS_IN_DEBUG` here, since right now disabling `OPEN_DEVTOOLS_IN_DEBUG` means there's no devtools shown at all.

r0x0r:
Nice work figuring out this one, it works like a charm. And you are correct about pointing out the OPEN_DEVTOOLS_IN_DEBUG  logic bug

PR #1663, opened 2025-06-22
[qt] Fix Qt parent methods being shadowed by custom attributes

### LINKED ISSUE #1660: Error on startup with `frameless=True` when using PySide6 / PyQt6
👋 I'm attempting to use qt as a renderer on Windows 11 24H2 (26100.4349), Python 3.12.0. On startup when using `frameless=True` (testing with `examples/transparent.py`), I get:
```
  File "C:\Users\Chris\Projects\pywebview\webview\platforms\qt.py", line 202, in eventFilter
    if object.parent() == self:
       ^^^^^^^^^^^^^^^
TypeError: 'BrowserView' object is not callable
```

I can reproduce this error across a few different versions of pywebview:
- `master` (currently `8e88814`)
- `5.4`
- `6.0` (currently `2bb808f`)

Though interestingly, the error is only fatal on PyQt6. For PySide6 (with the change from #1659 applied), the error is raised several times but it's never fatal - the application renders as expected.

Taking a look at where the error occurs, I can see most `object` instances do have a defined `parent` method to call:
Image

but when the `object` is a `Qt.WebPage`, `object.parent` _itself_ is an instance, not a method:
Image

It looks like this may be due to the upstream Qt `parent` method being shadowed by pywebview's custom `WebPage`:
https://github.com/r0x0r/pywebview/blob/2bb808f64576a86b0b5a0bb939e78ab76e49af0d/webview/platforms/qt.py#L248-L250

Simply removing the `self.parent` assignment fixes the issue - seemingly nothing downstream reads `self.parent`, instead always reading the `parent` argument directly in `__init__`. Please double check me on:
- if there _are_ any side-effects from this
- if any other classes need the same sort of fix

chrissantamaria:
Fixes #1660

Merging into `6.0` since I'm personally working off this branch, but happy to backport to `master` if that's preferable

### DIFF
--- a/webview/platforms/qt.py
+++ b/webview/platforms/qt.py
@@ -235,19 +235,18 @@ class BrowserView(QMainWindow):
     class NavigationHandler(QWebPage):
         def __init__(self, page):
             super().__init__(page.profile())
-            self.parent = page.parent
+            self._parent = page.parent
 
         def acceptNavigationRequest(self, url, type, is_main_frame):
             if settings['OPEN_EXTERNAL_LINKS_IN_BROWSER']:
                 webbrowser.open(url.toString(), 2, True)
                 return False
             else:
-                self.parent.load_url(url.toString())
+                self._parent.load_url(url.toString())
                 return True
 
     class WebPage(QWebPage):
         def __init__(self, parent=None, profile=None):
-            self.parent = parent
             if is_webengine and profile:
                 super(BrowserView.WebPage, self).__init__(profile, parent.webview)
             else:


webview/platforms/qt.py
```diff
@@ -235,19 +235,18 @@ def interceptRequest(self, info):
...
-            self.parent = page.parent
+            self._parent = page.parent
 
         def acceptNavigationRequest(self, url, type, is_main_frame):
             if settings['OPEN_EXTERNAL_LINKS_IN_BROWSER']:
                 webbrowser.open(url.toString(), 2, True)
                 return False
             else:
-                self.parent.load_url(url.toString())
+                self._parent.load_url(url.toString())
```
chrissantamaria, line 245: I haven't observed this breaking anything, but it seemed susceptible to the same issue so I eagerly fixed. Happy to revert.

webview/platforms/qt.py
```diff
@@ -235,19 +235,18 @@ def interceptRequest(self, info):
...
                 return True
 
     class WebPage(QWebPage):
         def __init__(self, parent=None, profile=None):
-            self.parent = parent
```
chrissantamaria, line 250: This appears to be unused downstream, just opted to remove instead of renaming

PR #1675, opened 2025-07-13
🛠️ Improve `cryptography` Dependency & PyPI Metadata

tutosrive:
> [!IMPORTANT]  
> I used `pip install pywebview` to install this _"module"/library_

I'm making this pull request for two reasons.

1. When testing the [HTTP Server example](https://pywebview.flowrl.com/guide/usage.html#http-server), I ran:
 ```shell
   (.venv) D:\CODE\PYTHON\pywebview\test1>python app5.py
     File "D:\CODE\PYTHON\pywebview\test1\app5.py", line 4, in <module>
       webview.start(ssl=True)
     File "D:\CODE\PYTHON\pywebview\test1\.venv\Lib\site-packages\webview\__init__.py", line 181, in start
       keyfile, certfile = __generate_ssl_cert()
                           ^^^^^^^^^^^^^^^^^^^^
     File "D:\CODE\PYTHON\pywebview\test1\.venv\Lib\site-packages\webview\__init__.py", line 346, in __generate_ssl_cert
       from cryptography import x509
   ModuleNotFoundError: No module named 'cryptography'
````

The simple fix was:

```bash
pip install cryptography
```

It would be helpful to mention this dependency (e.g. via `pywebview[ssl]`) in the guide and in `pyproject.toml`.

2. The use of the `ssl` extra is not explicitly mentioned in the example presented on the website.
   Since the example includes `ssl=True`, it might be clearer to show the full installation command as:

   ```bash
   pip install pywebview
   ```

   Therefore, I suggest moving `cryptography` from an optional extra to a main dependency, since it's needed by a default example.

---

Additionally, as an external contributor:

* I added a `README.md` entry in `pyproject.toml` so the full project description appears on PyPI.
* I included **Python 3.12** in the `classifiers` section of `pyproject.toml`, since I tested it successfully and it wasn’t listed originally.

Thank you for considering these suggestions!

— @tutosrive

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "pywebview"
 authors = [{ name = "Roman Sirokov", email = "roman@flowrl.com" }]
+readme = { file = "README.md", content-type = "text/markdown", charset = "utf-8" }
 description = "Build GUI for your Python program with JavaScript, HTML, and CSS"
 keywords = ["gui", "webkit", "html", "web"]
 requires-python = ">=3.7"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ classifiers = [
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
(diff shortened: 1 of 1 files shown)


r0x0r:
The reason why it is optional is difficulties in installing this package on some platforms. I don't know if the situation has got any better these days, but I do not wish for the installation to fail because of `cryptography`. Documentation should be updated though

PR #1682, opened 2025-08-01
Add dispatch_custom_event method to allow JS event communication from Python

S4mC:
This PR adds a dispatch_custom_event method to the window module.
It allows Python code to trigger custom JavaScript events, optionally passing data as event properties.
The JS event is created with { cancelable: true } to support event cancellation detection.

Useful for interacting with JS UIs without relying on evaluate_js alone..

### DIFF
--- a/examples/dispatch_custom_event.py
+++ b/examples/dispatch_custom_event.py
@@ -1,0 +1,31 @@
+""" Dispatch customs events in js:
+        window.addEventListener("personalEvent", (e) => {
+            console.log("Received from Python:", e.customvalue);
+        });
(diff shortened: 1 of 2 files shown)


r0x0r:
I am not sure about this one. This is essentially a wrapper around js code. If dispatching custom DOM events to be implemented,  it would be nice to have it both ways.

S4mC:
> I am not sure about this one. This is essentially a wrapper around js code. If dispatching custom DOM events to be implemented, it would be nice to have it both ways.

I created a new PR in which I implement that

PR #1683, opened 2025-08-02
Add before_closing event to support evaluate_js before window closes

S4mC:
Added `before_closing` event, which is executed before the closing event.

This event allows the use of `evaluate_js` to perform actions before the program exits. 
It does not interfere with the standard `closing` event and is especially useful in 
applications that already implement a closing handler but cannot use `evaluate_js` there.

If the handler returns `False`, the closing action is aborted. This allows users to 
prevent accidental exits, show confirmation dialogs, or perform pre-close validations.

**Example usage:**

```
def before_closing_function(arg1, arg2):
    print("before closing", arg1, arg2)

    result = window.evaluate_js('confirm("Are you very sure you want to close the window?");')
    
    if result:
        # If user confirms, show final message and close
        window.evaluate_js("""document.body.innerHTML += "Don't leave :("; alert("The window will now close");""")
    else:
        return False # Cancel closing the window
```

**And `after window = webview.create_window..`:**
`window.events.before_closing = lambda: before_closing_function("value1", "value2")`

### DIFF
--- a/examples/window.events.before_closing.py
+++ b/examples/window.events.before_closing.py
@@ -1,0 +1,25 @@
+import webview
+
+def on_closing():
+    # The on_closing event is executed when the window attempts to close, even if before_closing is running
+    if (not window.before_closing_active): # This filter prevents that from happening
+        print("on_closing")
+
+def before_closing_function(arg1, arg2):
+    print("before closing", arg1, arg2)
+
+    result = window.evaluate_js('confirm("Are you very sure you want to close the window?");')
+    
+    if result:
+        # If user confirms, show final message and close
+        window.evaluate_js("""document.body.innerHTML += "Don't leave :("; alert("The window will now close");""")
+    else:
+        return False # Cancel closing the window
+
+if __name__ == "__main__":
+    window = webview.create_window("Mi App", html="<html><head><body></body></head></html>")
+    
+    window.events.closing += on_closing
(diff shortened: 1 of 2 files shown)


S4mC:
> Cannot you use `evaluate_js` in closing event?

I tried using evaluate_js inside the closing event in Windows Forms, but the window would hang and not finish closing properly. To fix this (because I'm making a program that need that), I created that new event (before_closing) to handle the logic earlier.

S4mC:
> Cannot you use `evaluate_js` in closing event?

So evaluation_js should work in the closing event? I tried it, but it didn't work for me. I asked an AI, and it said it's not good practice to use evaluation_js in the closing event, and sometimes it doesn't work, so I added that new event to handle it.

S4mC:
> My take is that it should work in the closing event, but I have not tested it. I would not trust LLMs on this one, as this behaviour is not documented.

This is what happens to me when I try to use evaluated_js in the closing event, it is similar to the example I created in the pr for the before_closing event but only using the closing event.
https://github.com/user-attachments/assets/badf7894-5383-424b-aea8-a232b78255d4

r0x0r:
I have attempted to fix this, but I could not find an immediate solution. I have opened [a new ticket ](https://github.com/r0x0r/pywebview/issues/1699) for this and I am gonna close this pr now.

PR #1685, opened 2025-08-04
Add setting ALLOW_OPEN_DEVTOOLS

jeanbaptiste444:
Add 'ALLOW_OPEN_DEVTOOLS' setting to allow user to open devtools even if debug is False.


r0x0r:
What is the use case for this?

r0x0r:
This is a valid concern. I have modified the logging logic so that PYWEBVIEW_LOG takes precedence over the logging level even in the debug mode. The change can be found in the 6.0 branch.

PR #1700, opened 2025-08-29
[qt] Properly handle SIGINT to quit program

### LINKED ISSUE #1686: Unable to Intercept KeyboardInterrupt (Ctrl+C) on Main Thread with Qt Backend
**Environment Details**
- Operating System: Ubuntu 20.04  
- Python Version: 3.8.10  
- pywebview Version: 5.4  
- Python Qt Wrapper & Version:  PyQt6 6.7.1  

Hello! I am trying to create a pywebview application that can be closed cleanly from the terminal using a KeyboardInterrupt (Ctrl+C). However, when the webview.start() event loop is running on the main thread with the Qt backend, all attempts to catch SIGINT are ignored, and the application cannot be closed gracefully. It seems to be caught eventually but doesn't propogate up to the rest of the program that i have written.
This is the error that is shown, when this occurs:

```
  File "/home/bart/.local/lib/python3.8/site-packages/webview/platforms/qt.py", line 588, in eventFilter
    def eventFilter(self, object, event):
KeyboardInterrupt
```
this already happens with the hello world example

```python
import webview
webview.create_window('Hello world', 'https://pywebview.flowrl.com/')
webview.start(gui='qt')
```

is there any way to get this to work?

chrissantamaria:
Fixes #1686

### DIFF
--- a/webview/platforms/qt.py
+++ b/webview/platforms/qt.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import platform
+import signal
 import socket
 import sys
 import webbrowser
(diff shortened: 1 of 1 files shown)


webview/platforms/qt.py
```diff
@@ -908,6 +926,15 @@ def _create():
 
         _app = QApplication.instance() or QApplication(sys.argv)
 
+        signal.signal(signal.SIGINT, _sigint_handler)
+
+        # Create a timer to periodically allow the Python interpreter to run
+        # This enables the signal handler to be called even when the Qt event loop is running
+        # The timer just calls a no-op function to trigger Python execution
+        timer = QtCore.QTimer()
+        timer.start(500)
```
chrissantamaria, line 935: This is arbitrary, happy to change to whatever feels appropriate

PR #1711, opened 2025-09-09
Add can_create_directories parameter to create_file_dialog for macOS

### LINKED ISSUE #1710: Add support for 'New Folder' button in FOLDER_DIALOG on macOS (canCreateDirectories)
### Description

Currently, when using `webview.FileDialog.FOLDER` (or the older `webview.FOLDER_DIALOG`) on macOS, the native folder selection dialog doesn't show a "New Folder" button, preventing users from creating new directories during the selection process. This limits the user experience when an application needs users to select or create a destination folder.

### Current Behavior

Looking at the https://github.com/r0x0r/pywebview/blob/master/webview/platforms/cocoa.py, the create_file_dialog method for FOLDER_DIALOG uses NSOpenPanel and sets:

`open_dlg.setCanChooseDirectories_(dialog_type == FileDialog.FOLDER)`

However, it does not set the canCreateDirectories property, which means the "New Folder" button is not shown in the dialog.

### Expected Behavior

The folder dialog should display a "New Folder" button (typically in the bottom-left corner on macOS), allowing users to create new directories without leaving the dialog.

### Technical Details

On macOS, `NSOpenPanel` inherits from `NSSavePanel`, which provides the `canCreateDirectories` property. According to Apple's documentation, setting this property to YES enables the "New Folder" button:

```
NSOpenPanel *panel = [NSOpenPanel openPanel];
[panel setCanChooseFiles:NO];
[panel setCanChooseDirectories:YES];
[panel setCanCreateDirectories:YES];  // This enables the "New Folder" button

In PyObjC (which PyWebView uses), this would be:
open_dlg.setCanCreateDirectories_(True)
```

### Proposed Solution

The simplest fix would be to add this line in the `create_file_dialog` method in `cocoa.py` when handling `FOLDER_DIALOG`:

```
if dialog_type == FileDialog.FOLDER:
    open_dlg.setCanChooseDirectories_(True)
    open_dlg.setCanCreateDirectories_(True)  # Add this line

```
Alternatively, if you want to make it configurable, add an optional parameter:

```
def create_file_dialog(
    dialog_type=FileDialog.OPEN,
    directory='',
    allow_multiple=False,
    save_filename='',
    file_types=(),
    can_create_directories=None  # New parameter
):
    # ...
    if dialog_type == FileDialog.FOLDER:
        open_dlg.setCanChooseDirectories_(True)
        # Default to True for folder dialogs, but allow override
        if can_create_directories is not None:
            open_dlg.setCanCreateDirectories_(can_create_directories)
        else:
            open_dlg.setCanCreateDirectories_(True)
```

This could be enabled by default for folder dialogs, as it's a common user expectation on macOS.

### Use Case

This feature is particularly useful for applications that:
- Need users to select output/export directories
- Manage project files in user-specified locations
- Require users to organize content in new folders
- Import/export workflows where users often create new folders for organization

### Environment

- PyWebView version: 6.0+
- Platform: macOS (though similar functionality could be useful on other platforms)

### Workaround

Currently, developers must either:
1. Use SAVE_DIALOG (which allows folder creation) and extract the directory path
2. Implement a two-step process: select parent folder, then programmatically create subdirectories
3. Build custom folder creation UI outside of the native dialog

None of these workarounds provide the seamless user experience of the native "New Folder" button.

### Related Issues

- #971 - Add support for Open Directory mode (addressed basic folder selection but not folder creation)
- #578 - Save Dialog issues on macOS (different but related dialog functionality)

### Additional Context

Similar functionality exists on other platforms:
- Windows folder dialogs often include a "New Folder" button by default
- Linux file managers typically allow folder creation in dialogs

This would improve feature parity across platforms and enhance the user experience for macOS users.

camdarley:
## Summary
- Adds support for the "New Folder" button in macOS folder dialogs
- Adds optional `can_create_directories` parameter to `create_file_dialog` method
- Parameter defaults to `True` to maintain backward compatibility

## Changes
- Added `can_create_directories` parameter to `Window.create_file_dialog()` in `webview/window.py`
- Added parameter support in macOS platform implementation (`webview/platforms/cocoa.py`)
- Sets `NSOpenPanel.setCanCreateDirectories_()` when dialog type is `FOLDER`

## Testing
Tested on macOS with a test script that verifies:
- Default behavior (can_create_directories=True) shows "New Folder" button
- Setting can_create_directories=False hides "New Folder" button
- Explicitly setting can_create_directories=True shows "New Folder" button

## Related Issue
Fixes #1710

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -901,7 +901,7 @@ class BrowserView:
         return JSResult.result
 
     def create_file_dialog(
-        self, dialog_type, directory, allow_multiple, save_filename, file_filter, can_create_directories, main_thread=False
+        self, dialog_type, directory, allow_multiple, save_filename, file_filter, main_thread=False
     ):
         def create_dialog(*args):
             dialog_type = args[0]
@@ -934,8 +934,8 @@ class BrowserView:
                 open_dlg.setCanChooseDirectories_(dialog_type == FileDialog.FOLDER)
 
                 # Enable creating new folders in folder dialogs (macOS)
-                if dialog_type == FileDialog.FOLDER and can_create_directories:
-                    open_dlg.setCanCreateDirectories_(can_create_directories)
+                if dialog_type == FileDialog.FOLDER:
+                    open_dlg.setCanCreateDirectories_(True)
 
                 # Enable / disable multiple selection
                 open_dlg.setAllowsMultipleSelection_(allow_multiple)
@@ -1384,7 +1384,7 @@ def create_confirmation_dialog(title, message, uid):
     return result
 
 
-def create_file_dialog(dialog_type, directory, allow_multiple, save_filename, file_types, can_create_directories, uid):
+def create_file_dialog(dialog_type, directory, allow_multiple, save_filename, file_types, uid):
     file_filter = []
 
     # Parse file_types to obtain allowed file extensions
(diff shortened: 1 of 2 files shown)


r0x0r:
This could be enabled by default without the can_create_directories option

PR #1740, opened 2025-10-25
qt: support REMOTE_DEBUGGING_PORT

cybergrind:
This adds support `REMOTE_DEBUGGING_PORT` to the qt platform

I have tested it under qt5 - it is working (but have some internal qt5 bug when you're trying to type into input fields, otherwise everything seems working ok)
qt6 - working as expected


r0x0r:
Hey looks good, but could you fix the formatting? I have introduced ruff formatter with pre-commit hook that takes care of formatting automatically. More info on it can be found here https://pywebview.flowrl.com/contributing/development.html#environment-set-up

PR #1741, opened 2025-10-25
linux: kali needs no-sandbox flag

cybergrind:
Kali linux also needs `--no-sandbox` treatment


cybergrind:
Actually, it looks like an issue with pyinstaller, rather than pywebview. I was able to launch it without this patch from source code, but cannot run built version and it requires to pass `--no-sandbox` directly to the app itself.

I will reopen it if find something that actually fixes this behavior

PR #1743, opened 2025-10-28
Release retained Cocoa objects to fix “ghost” windows

AlecHaring:
After updating to macOS Tahoe, I noticed a weird bug with secondary windows. When I'd close them, they'd visually disappear, but I'd then run into issues where I couldn't click on certain parts of the screen. It turned out the windows weren't actually closing; they were just becoming invisible:

https://github.com/user-attachments/assets/f46915b4-d8c3-4ff1-8aa8-365f197363c3

## Root cause
Inside `BrowserView.__init__`, several Cocoa objects are explicitly retained (`NSWindow`, `WKWebView`, and delegates), but are never released. A closed-but-retained `NSWindow` is removed from the screen, but not deallocated. This results in a ghost window, as seen in the above video.

## Solution
Release the retained objects when the window is closed, inside `windowWillClose_`.

## Note
I'm honestly not entirely convinced explicitly retaining the objects in `__init__` is necessary to begin with, since `BrowserView` keeps strong Python references to these objects. Unless I'm missing something, `pyobjc` automatically handles the retain/release cycle behind the scenes ([source](https://github.com/ronaldoussoren/pyobjc/blob/main/docs/dev/structure.rst#reference-counts)).

I kept the init code as is, though, to preserve existing logic. The decision to switch away from explicit retains can be made at a future date.

### DIFF
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -108,6 +108,12 @@ class BrowserView:
             i.webview.removeFromSuperview()
             i.webview = None
 
+            # release everything we retained in __init__
+            i._browserDelegate.release()
+            i._windowDelegate.release()
+            i._appDelegate.release()
+            i.window.release()
+
             i.closed.set()
             if BrowserView.instances == {}:
                 BrowserView.app.stop_(self)


r0x0r:
Thanks for the fix. I am not sure about retaining either.

aDifferentJT:
Thanks for fixing this, I was running into the same issue on 6.1 and this has fixed it. Do you know when a 6.2 might come out that would include this fix? I'd prefer to use a stable dependency from pypi rather than depending on the git tree.

Thanks

PR #1665, opened 2025-06-26
Add Windows icon support to pywebview

gitcnd:
- Add user-provided icon loading in WinForms BrowserForm
- Support ICO file format for Windows window icons
- Update documentation to reflect Windows icon support
- Maintain fallback to application icon on failure

Fixes window icons not working on Windows platforms. Icons now work on Windows (WinForms), GTK, and QT.

Screenshot from testing:-
pic_2025-06-26_12 56 56_1780

**disclaimer** - this is claude4's work.  partial log is below.

---

# Fixing window icons in pywebview (2025-06-26 12:08:44)

_**User**_

this is a fork of the github pywebview project.

browse it 

we are going to attempt to address the issue of window icons not working on microsoft-windows platforms.

---

_**Assistant**_

---

# Response 1

I'll help you investigate the pywebview project and address the window icons issue on Microsoft Windows platforms. Let me start by exploring the project structure to understand the codebase.

---

<details>
            <summary>Listed current directory • **20** results</summary>
        
| Name |
|-------|
| 📄 `.gitignore` |
| 📄 `.cursorindexingignore` |
| 📁 `.specstory` |
| 📁 `.git` |
| 📁 `webview` |
| 📁 `tests` |
| 📄 `pyproject.toml` |
| 📄 `requirements.txt` |
| 📁 `logo` |
| 📁 `interop` |
| 📁 `examples` |
| 📁 `docs` |
| 📄 `appveyor.yml` |
| 📄 `MANIFEST.in` |
| 📄 `README.md` |
| 📄 `SECURITY.md` |
| 📄 `.pre-commit-config.yaml` |
| 📄 `LICENSE` |
| 📁 `.github` |
| 📄 `.editorconfig` |

</details>

---

<details>
            <summary>Listed directory webview • **16** results</summary>
        
| Name |
|-------|
| 📄 `window.py` |
| 📄 `screen.py` |
| 📄 `util.py` |
| 📄 `py.typed` |
| 📁 `platforms` |
| 📄 `menu.py` |
| 📄 `localization.py` |
| 📁 `lib` |
| 📁 `js` |
| 📄 `guilib.py` |
| 📄 `http.py` |
| 📄 `errors.py` |
| 📄 `event.py` |
| 📁 `dom` |
| 📁 `__pyinstaller` |
| 📄 `__init__.py` |

</details>

---

<details>
            <summary>Listed directory webview/platforms • **9** results</summary>
        
| Name |
|-------|
| 📄 `winforms.py` |
[...]

Issue #1411, opened 2024-06-15
How to replace the icon in the top left corner?


plane55:
### Specification
- pywebview version: 4.4.1
- operating system: windows 10
- web renderer: 

### Description

  How to replace the icon in the top left corner?

### Practicalities
![Snipaste_2024-06-15_14-30-15](https://github.com/r0x0r/pywebview/assets/41913958/7fcf6de5-5c5d-4e12-981c-10cd041a49d7)

r0x0r:
_pywebview_  extracts and displays application icon from the executable. In other words you specify the application icon via a bundler.

Cqy11223344:
Webview. start (gui='qt '), when the gui is configured as qt, the correct icon cannot be displayed. If it is cef or other parameters, it is normal

Cqy11223344:
![Uploading QQ五笔截图未命名.png…]()

r0x0r:
@Cqy11223344 Do you get this behaviour when application is frozen? The icon support comes only with frozen apps.

GbSeason:
This is my solution. Firstly, this library uses native window handles in the Windows system, so you need to consult the. NET documentation to find the loading method for window icons. Fortunately, there is just a method to load ICO files based on the path.

![image](https://github.com/user-attachments/assets/0fa5d44d-d85f-41ba-882c-ef8ae908dbc1)

At this point, you need to make slight changes to the source code by adding the path attribute of the ICO file to the custom window object, and adding code to specify the window icon when creating the window

![image](https://github.com/user-attachments/assets/c7b0b315-c072-4eee-a145-fe3fcd4ecd38)

![image](https://github.com/user-attachments/assets/01db6499-ffa6-4e9b-9fda-0a569ac73749)

![image](https://github.com/user-attachments/assets/11127a6d-e5ad-44ba-9313-6dcfcb14ebb1)

![image](https://github.com/user-attachments/assets/77c0ee66-ea97-446b-abf8-980f828f1e40)

Create a webview and set 'icon_path'.

`webview.create_window('Hello world', html="<div></div>", icon_path=f"{img}/favicon_play.ico")`

Now the icon is changed.

![image](https://github.com/user-attachments/assets/b23d0fd4-2de4-48fc-b13a-e98ba74cc240)

Cqy11223344:
I tried to modify it using the above method, but it still doesn't work. You can try calling it using the following method.
webview.start(gui='qt')

[...]

PR #1755, opened 2025-12-04
Optionally set icon in WinForms

paco-sevilla:
This PR allows to set an icon using `webview.start(icon=<file_path>)` when the platform is WinForms (i.e. default on Windows).

This is implemented in a similar fashion as for the [Qt](https://github.com/paco-sevilla/pywebview/blob/master/webview/platforms/qt.py#L499-L501) and [GTK](https://github.com/paco-sevilla/pywebview/blob/master/webview/platforms/gtk.py#L250-L252), with the difference that we check whether `_state['icon']` exists (and is a file).
This makes sure that the icon set when freezing an app can be used without having to bundle its file.
I hope this addresses the concerns raised in https://github.com/r0x0r/pywebview/pull/1665 and https://github.com/r0x0r/pywebview/issues/1411.

**Background/Rationale:** At the company where I work, we don't freeze our (internal) apps, but we deploy them to a Python Registry and install them with [`uv tools install`](https://docs.astral.sh/uv/guides/tools/#installing-tools). This allows us to deploy a single (pure-Python) package for both Windows and Linux users.
Even for people that freeze their apps, I think it would be nice if they could show the icon during development (on Windows).

### DIFF
--- a/webview/platforms/winforms.py
+++ b/webview/platforms/winforms.py
@@ -225,12 +225,15 @@ class BrowserView:
             self.old_state = self.WindowState
 
             # Application icon
-            handle = kernel32.GetModuleHandleW(None)
-            icon_handle = windll.shell32.ExtractIconW(handle, sys.executable, 0)
+            if _state['icon'] and os.path.isfile(_state['icon']):
+                self.Icon = Icon(_state['icon'])
+            else:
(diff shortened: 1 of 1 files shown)


r0x0r:
Fair enough, thanks for the contribution

PR #1794, opened 2026-03-23
feat: add hwnd property to Window

cnlnr:
Exposing the HWND enables advanced native manipulations, such as window transparency, blur effects, and custom styles via the Win32 API.

### DIFF
--- a/webview/platforms/winforms.py
+++ b/webview/platforms/winforms.py
@@ -1055,5 +1055,12 @@ def get_screens():
     return screens
 
 
+def get_hwnd(uid):
+    i = BrowserView.instances.get(uid)
+    if i:
+        return int(i.Handle.ToInt64())
+    return None
+
+
 def add_tls_cert(_):
     return
--- a/webview/window.py
+++ b/webview/window.py
@@ -225,6 +225,12 @@ class Window:
         return height
 
     @property
+    def hwnd(self) -> int | None:
+        if self.gui is not None:
+            return self.gui.get_hwnd(self.uid)
+        return None
+
+    @property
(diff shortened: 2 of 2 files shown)


cnlnr:
Below is an example of setting window transparency
``` python
import webview
import ctypes

# Pre-load user32.dll
user32 = ctypes.windll.user32

def on_shown(window):
    """
    This function triggers as soon as the window is visible.
    The hwnd is guaranteed to be available here.
    """
    hwnd = window.hwnd
    # 1. Enable layered window style
    style = user32.GetWindowLongW(hwnd, -20)
    user32.SetWindowLongW(hwnd, -20, style | 0x80000)
    
    # 2. Set transparency (Alpha: 180)
    user32.SetLayeredWindowAttributes(hwnd, 0, 180, 0x2)
    print(f"Effect applied to HWND: {hwnd}")

if __name__ == '__main__':
    window = webview.create_window(
        'Native Transparency', 
        html='<h1 style="font-family:sans-serif;">Window is Ready!</h1>'
    )

    # Hardcore approach: Bind to the 'shown' event
    # This ensures the code runs AFTER the window is created.
    window.events.shown += lambda: on_shown(window)

    webview.start()
```

r0x0r:
hwnd can be accessed already via `window.native.Handle`. See `https://pywebview.flowrl.com/api/#window-native`

Does your transparency solution work reliably with edgehtml?

cnlnr:
https://github.com/r0x0r/pywebview/issues/1779
Do you mean the issue where the empty space background isn't transparent?

PR #1797, opened 2026-03-26
Fix use-after-free crash in BrowserView.get_instance on macOS

smparkes:
On macOS ARM64, the app crashes with EXC_BREAKPOINT (SIGTRAP, pointer authentication trap) in object_getClass when a deferred Cocoa callback (performSelectorOnMainThread) fires after a window has been destroyed.

Root cause: get_instance iterates BrowserView.instances comparing each instance's .window attribute against a target NSWindow. When a window has been destroyed (windowWillClose_ releases the window), the .window attribute points to freed memory. Comparing it triggers object_getClass on the freed pointer, which fails the ARM64 pointer authentication check and kills the process.

The original code had two bugs:
1. Used bare getattr(i, attr) — crashes if the attribute value is a freed Cocoa object whose __eq__ triggers object_getClass
2. Caught only AttributeError and used break — stops searching on any missing attribute, missing valid instances later in the dict

Fix get_instance to:
- Use getattr(i, attr, None) with a None check to skip freed/cleared attributes without comparing them
- Catch all exceptions (not just AttributeError) during comparison
- Continue iterating instead of breaking, so later valid instances are still found

Also fix windowWillClose_ to:
- Clear the window delegate (setDelegate_(None)) before releasing, preventing further delegate callbacks on the freed window
- Guard against missing instance (if not i: return)
- Guard webview access (if i.webview:) before cleanup

And fix windowShouldClose_ to return YES if instance not found.


smparkes:
Hmmm ... not sure about

Comprehensive CI / macOS (pull_request)Cancelled after 20m

Timeout?

r0x0r:
Lint check fails, you need to reformat your code using precommit hook.

What are those Winforms changes?

smparkes:
I had some other things going on in this repo and I think the branch got combined and pushed.

The winform changes should be on another branch. Described in https://github.com/r0x0r/pywebview/pull/1797/changes/c988ea467157096131564db5527905d7fd3f2356
Critical for Windows ARM64 because various DLLs are not provided for ARM64. I'm happy to contribute that but I don't know that I have a good grasp of the compatibility implications ... like continue to support .NET 4 as well as .NET 6/8 etc. so I don't know how acceptible they are.

I've had a few more crashes related to cocoa stuff and I've got logging now so I know what sequences of my app are triggering. I'll try to provide more PRs if I have repeatable failures not attributable to my bugs.

smparkes:
Dug deeper. I'm going to close this one in favor of some others soon.

PR #1799, opened 2026-03-30
Fix use-after-free crash on macOS ARM64 by disabling releasedWhenClosed

smparkes:
On macOS ARM64, pywebview crashes with EXC_BREAKPOINT (SIGTRAP, pointer authentication trap) when a Cocoa callback accesses a window that was freed after closing. The crash is demonstrated by running the existing test suite on macOS 26.4 ARM64, where it reliably crashes around test_multi_window::test_load_url.

Root cause: NSWindow defaults to releasedWhenClosed=YES, which makes Cocoa release the window when it closes. BrowserView also manually calls .retain() in __init__ and .release() in windowWillClose_ — a balanced pair. But the releasedWhenClosed auto-release is untracked, creating an over-release: the window's retain count eventually reaches zero while internal Cocoa/WebKit references still exist. When a stale callback (e.g. from WebKit's async IPC) later accesses the freed window, object_getClass hits a corrupted ISA pointer and triggers an ARM64 pointer authentication trap.

Confirmed via lldb: breakpoints on [NSWindow
initWithContentRect:styleMask:backing:defer:] and [NSWindow dealloc] show the window being deallocated twice at the same address with no intervening allocation — a confirmed over-release, not address reuse. The crash address matches the double-deallocated window. The new test_no_double_dealloc_on_window_close test reproduces this by running multi-window tests under lldb and verifying no address is deallocated without a preceding init.

Fix: call setReleasedWhenClosed_(False) after window creation. This removes the untracked release, keeping the retain/release accounting balanced between pywebview's .retain()/.release() and pyobjc's proxy lifecycle.


r0x0r:
Thanks for fixing this. I removed a test as I try to keep tests focusing on functional features

PR #1805, opened 2026-03-31
fix transparency

Mnaisuka:
https://github.com/r0x0r/pywebview/issues/1779#issuecomment-4150520275

### DIFF
--- a/webview/platforms/winforms.py
+++ b/webview/platforms/winforms.py
@@ -756,11 +756,15 @@ def create_window(window):
             browser.Hide()
             browser.Opacity = 1
         elif window.transparent and is_chromium:
-            # hack to make transparent window work
-            # window is started hidden and shown on Navigating event.
-            # no idea why this works
-            browser.Show()
-            browser.Hide()
+            def on_shown():
+                _user32 = ctypes.windll.user32
+                _hwnd = window.native.Handle.ToInt32()
+                
+                _style = _user32.GetWindowLongW(_hwnd,-20)
+                _user32.SetWindowLongW(_hwnd,-20,_style | 0x80000)
+
+                _user32.SetLayeredWindowAttributes(_hwnd,0,180,0x2)
+            window.events.shown  += on_shown
         else:
             browser.Show()
 
--- a/webview/platforms/winforms.py
+++ b/webview/platforms/winforms.py
@@ -758,7 +758,7 @@ def create_window(window):
         elif window.transparent and is_chromium:
             browser.Opacity = 0
 
-            def on_loaded(sender=None, args=None):
+            def on_loaded(window):
                 browser.Opacity = 1
                 window.hide()
                 window.show()
--- a/webview/platforms/winforms.py
+++ b/webview/platforms/winforms.py
@@ -756,15 +756,14 @@ def create_window(window):
             browser.Hide()
             browser.Opacity = 1
         elif window.transparent and is_chromium:
-            def on_shown():
-                _user32 = ctypes.windll.user32
-                _hwnd = window.native.Handle.ToInt32()
-                
-                _style = _user32.GetWindowLongW(_hwnd,-20)
-                _user32.SetWindowLongW(_hwnd,-20,_style | 0x80000)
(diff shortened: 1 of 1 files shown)


Mnaisuka:
If you set `transparent` and then set `hidden` to `true`, `transparent` will not take effect, but I'm not sure if this is a bug.

```python
if window.hidden:
    browser.Opacity = 0
    browser.Show()
    browser.Hide()
    browser.Opacity = 1
elif window.transparent and is_chromium:
    # hack to make transparent window work
    # window is started hidden and shown on Navigating event.
    # no idea why this works
    browser.Show()
    browser.Hide()
else:
    browser.Show()
```
```python
if window.hidden and not window.transparent:
    browser.Opacity = 0
    browser.Show()
    browser.Hide()
    browser.Opacity = 1
elif window.transparent and is_chromium:
    def on_shown():
        _user32 = ctypes.windll.user32
        _hwnd = window.native.Handle.ToInt32()
        
        _style = _user32.GetWindowLongW(_hwnd,-20)
        _user32.SetWindowLongW(_hwnd,-20,_style | 0x80000)

        _user32.SetLayeredWindowAttributes(_hwnd,0,180,0x2)
        
        if window.hidden:
            _user32.ShowWindow(_hwnd, 0)

    window.events.shown  += on_shown
else:
    browser.Show()
```

r0x0r:
Can you fix lint and format job?

r0x0r:
Transparency does not work properly with `examples/transparent.py`. It looks like this for me 
image

Mnaisuka:
Perhaps refreshing the window in the loaded event would solve the problem?
`2GAFQBC21F%T3II10 _97

r0x0r:
This works for me, but the problem is that the existing solution works for me as well. Maybe someone for whom current solution does not cut could test it.

Cristian-F-M:
### 🧪 Windows Transparency Test Results

Hi, I've tested the proposed changes for transparency on Windows (Chromium/WebView2) and found that the current implementation in the PR has some limitations. Below are the results of two different approaches:

#### ❌ Test A: Testing the "Show/Hide" Hack
I tested the approach of using only `browser.Show()` / `Hide()` (as seen in some workarounds), but it results in a **solid background** on my machine, failing to achieve true transparency.


```python
if window.hidden:
    browser.Opacity = 0
    browser.Show()
    browser.Hide()
    browser.Opacity = 1
elif window.transparent and is_chromium:
    # This hack is not enough for consistent transparency
    browser.Show()
    browser.Hide()
else:
    browser.Show()
```
### It looks like:

image

#### ✅ Test B: Testing Win32 API Layered Window
I also tested forcing the `WS_EX_LAYERED` style via `ctypes` and the Win32 API. In this case, the transparency works perfectly as intended. This confirms that manipulating the window styles at the OS level seems necessary for Chromium to render transparency correctly on Windows.

```python
if window.hidden and not window.transparent:
    browser.Opacity = 0
    browser.Show()
    browser.Hide()
    browser.Opacity = 1
elif window.transparent and is_chromium:
    def on_shown():
        import ctypes
        _user32 = ctypes.windll.user32
        _hwnd = window.native.Handle.ToInt32()
        
        # GWL_EXSTYLE = -20 | WS_EX_LAYERED = 0x80000
        _style = _user32.GetWindowLongW(_hwnd, -20)
        _user32.SetWindowLongW(_hwnd, -20, _style | 0x80000)

        # LWA_ALPHA = 0x2
        _user32.SetLayeredWindowAttributes(_hwnd, 0, 180, 0x2)
        
        if window.hidden:
            _user32.ShowWindow(_hwnd, 0)

    window.events.shown += on_shown
else:
    browser.Show()
```
### It looks like:
Captura de pantalla 2026-04-07 144347

cnlnr:
I suggest you take a look at how VS Code implements this.

Mnaisuka:
The current solution seems to still have some new issues. When the transparent window is forcibly minimized, the transparency becomes ineffective—for example, when using Windows Peek or pressing Win + D.

PR #1808, opened 2026-04-08
Fix EdgeChromium fallback when NavigationCompleted is not fired after load_url

orztrickster:
In some cases, after calling load_url, on_navigation_completed is not triggered automatically.
When this happens, self.events._pywebviewready.clear() is never followed by window.events._pywebviewready.set().
As a result, subsequent API calls can time out and raise Main window failed to start after 20 seconds.
To address this, I suggest adding a safeguard: if on_navigation_completed has not been invoked within 15 seconds after load_url, trigger the completion handler as a fallback.


r0x0r:
It is a sound idea, but I am not a fan of this implementation. Some comments
- Timeout should be configurable via `webview.settings["LOAD_TIMEOUT"]` or something similar
- Firing a load event on timeout is confusing. Maybe a better approach is to introduce another event eg `load_timeout` and resolve when either `loaded` or `load_timeout` fires.
- This pr handles only edgechromium, other platforms need to be taken into account as well

orztrickster:
Added a _load_timeout flag. Calling load_url sets _load_timeout to True. If on_navigation_completed runs afterwards, it sets _load_timeout back to False. When _api_call checks and finds that _load_timeout is still True after 20 seconds, it knows on_navigation_completed was never called. In that case, instead of raising 'Main window failed to start', it calls inject_pywebview as a fallback. The _load_timeout flag is also added to the other platforms.

PR #1828, opened 2026-07-25
[Cocoa] fix window coordinates on multi-monitor setups

### LINKED ISSUE #1820: Frameless window drag jumps off-screen on macOS multi-monitor setups (cocoa backend)
**Summary**

On macOS, a frameless pywebview window dragged via the bundled pywebview-drag-region JS jumps far off the visible workspace mid-drag when the user has multiple monitors arranged so the active window's cached screen has a non-zero origin (e.g. a secondary monitor positioned to the LEFT of the primary, at screen X = -1920). Single-monitor setups are unaffected.

The cause is a coordinate-system double-count between the bundled drag JS and BrowserView.move in the cocoa backend.

**Environment**

pywebview: 6.2.1 (also present in 6.1, 6.2)
Backend: cocoa (webview/platforms/cocoa.py)
Python: 3.14 (also reproducible on 3.11+)
macOS: 14 / 15 (Sonoma / Sequoia)
Hardware: Apple Silicon, Retina display + a secondary external monitor positioned to the left of the primary in System Settings → Displays.
Reproduction
Arrange displays so a secondary monitor sits to the LEFT of the primary (System Settings → Displays → drag the secondary onto the left side). Verify the secondary now has negative screen-X coordinates.
Create a minimal frameless window with a drag region:
import webview

```
HTML = """
<!doctype html>
<html><body style="margin:0">
  <div class="pywebview-drag-region" style="height:40px;background:#ccc">drag me</div>
  <p style="padding:1em">repro</p>
</body></html>
"""

window = webview.create_window("repro", html=HTML, frameless=True,
                                 easy_drag=False, width=1280, height=800)
webview.start()
```

Move the window onto the secondary (left) monitor.
Grab the drag-region with the mouse and start dragging.
Expected: the window tracks the cursor 1:1.

Actual: on the first mousemove, the window jumps far to the left, often beyond the edge of the secondary monitor's bounds, becoming difficult or impossible to recover without Mission Control or a full restart.

**Root cause**

The drag is handled by JS in webview/js/customize.js, which on each mousemove sends absolute screen coordinates to Python:

```
// webview/js/customize.js
function onMove(ev) {
    var x = ev.screenX - initialX;   // absolute screen X of the window's left edge
    var y = ev.screenY - initialY;
    window.pywebview._jsApiCallback('pywebviewMoveWindow', [x, y], 'move');
}
```

x is the window's absolute screen-X position (browser-convention, can be negative on a left-side secondary monitor).

The cocoa backend's BrowserView.move then ADDS self.screen.origin.x to that x value before calling setFrameTopLeftPoint::

```
# webview/platforms/cocoa.py:811-815
def move(self, x, y):
    flipped_y = self.screen.size.height - y
    self.window.setFrameTopLeftPoint_(
        AppKit.NSPoint(self.screen.origin.x + x, self.screen.origin.y + flipped_y)
    )
```

self.screen is cached from NSScreen.mainScreen().frame() at window creation. On single-monitor setups its origin is (0, 0) and the addition is a no-op — which is why the bug doesn't surface there. On multi-monitor setups where the cached screen has non-zero origin (or the window has migrated to a screen with a different origin than the cached one), the addition double-counts:

JS sends x = -1600 (the window's actual screen X on the secondary monitor)
Python computes screen.origin.x + x = (-1920) + (-1600) = -3520
Window jumps to x = -3520 — well off the visible workspace.
The same problem exists on the Y axis for vertical multi-monitor layouts (screen.origin.y != 0), though we've only observed/confirmed the X variant.

**Suggested fix**

Drop the screen.origin.x addition. The JS already sends absolute screen coordinates; Python should pass them through:

```
 def move(self, x, y):
-    flipped_y = self.screen.size.height - y
+    primary = AppKit.NSScreen.mainScreen().frame()
+    flipped_y = primary.size.height - y
     self.window.setFrameTopLeftPoint_(
-        AppKit.NSPoint(self.screen.origin.x + x, self.screen.origin.y + flipped_y)
+        AppKit.NSPoint(x, flipped_y)
     )
```

The flipped_y change uses the primary screen's height as the Y reference, matching the browser-convention Y the JS sends (which is also measured from the top of the primary screen). This makes Y correct for vertical multi-monitor layouts as well.

On single-monitor: screen.origin = (0, 0) and the new code is identical to the old, so no regression.

**Workaround (downstream)**

For downstream projects, monkey-patching BrowserView.move at startup is effective:

```
import sys

def _patch_pywebview_multi_monitor_drag():
    if sys.platform != "darwin":
        return
    try:
        from webview.platforms.cocoa import BrowserView
        import AppKit
    except ImportError:
        return

    def _patched_move(self, x, y):
        flipped_y = self.screen.size.height - y
        self.window.setFrameTopLeftPoint_(
            AppKit.NSPoint(x, self.screen.origin.y + flipped_y)
        )

    BrowserView.move = _patched_move

# Call before `webview.create_window` / `webview.start`.
_patch_pywebview_multi_monitor_drag()
```

Happy to send a PR if the fix shape looks right.

VictorPurMar:
Fixes #1820.

`move()` added the cached screen's origin to coordinates that are already absolute. The drag region JS sends `ev.screenX / ev.screenY`, and `get_position` already returned absolute x, so the two disagreed. With one monitor the origin is (0, 0) and the bug is invisible.

All in `cocoa.py`:

- `move()`: pass x through, flip y against the primary screen instead of the cached one.
- `get_position()` and the `moved` event: flip y against the primary screen too, so a move round-trips. Both used the cached screen height without its origin, wrong on any screen whose origin is not (0, 0).
- initial placement: add the target screen's origin at the call site, so `create_window(screen=..., x=..., y=...)` stays relative to that screen. Same split as qt (`qt.py:490` vs `qt.py:686`).
- `_primary_screen_height()`: `NSScreen.mainScreen()` is the screen holding the key window, not the primary one, so it cannot be used for the flip. The primary is `NSScreen.screens()[0]`, which is what WebKit flips `screenX / screenY` against.

Dropped the `Failed to obtain screen` guard in `get_position`, it no longer reads `i.screen`.

## Testing

macOS 26.5, Python 3.13, primary 1728x1117 at (0, 0), secondary 1920x1080 at (-1920, 213). Window created on the secondary monitor, then moved with the absolute coordinates the drag JS sends:

| | before | after | expected |
| --- | --- | --- | --- |
| `move(-1720, 124)` landed at | (-2280, -52) | (-1720, 124) | (-1720, 124) |
| `moved` event payload | (-2280, -89) | (-1720, 124) | (-1720, 124) |
| `window.x` / `window.y` after `x=100, y=100` | (-1820, -113) | (-1820, -76) | (-1820, -76) |

y is off by 176, the secondary screen's offset from the primary's top, so the y axis is affected as well. The raw x double count is -3640, clamped to -2280 by AppKit's on-screen constraint. The 37px reporting error is the primary height minus the cached screen height.

`center()` is unaffected: identical AppKit origin before and after, only its reported position changes, from (-1160, 177) to (-1160, 214), and 214 is the correct value.

`pytest tests/test_move_window.py tests/test_screens.py tests/test_multi_window.py tests/test_frameless.py tests/test_fullscreen.py tests/test_state.py tests/test_start.py`: 36 passed, with one monitor and with two.

gtk adds the screen origin the same way at `gtk.py:534`. Left alone, I have no way to test it.

### DIFF
--- a/webview/platforms/cocoa.py
+++ b/webview/platforms/cocoa.py
@@ -53,6 +53,12 @@ logger.debug('Using Cocoa')
 renderer = 'wkwebview'
 
 
+def _primary_screen_height():
+    """Height of the primary screen, ie. screens()[0], not the focused mainScreen()."""
+    screens = AppKit.NSScreen.screens()
+    return screens[0].frame().size.height if screens else 0
+
+
 class BrowserView:
     instances = {}
     app = AppKit.NSApplication.sharedApplication()
@@ -151,8 +157,7 @@ class BrowserView:
             i = BrowserView.get_instance('window', notification.object())
             if i:
                 frame = i.window.frame()
-                screen = i.window.screen().frame()
-                flipped_y = screen.size.height - frame.size.height - frame.origin.y
+                flipped_y = _primary_screen_height() - frame.origin.y - frame.size.height
                 i.pywebview_window.events.moved.set(frame.origin.x, flipped_y)
 
     class JSBridge(AppKit.NSObject):
@@ -667,7 +672,9 @@ class BrowserView:
         self.window.setFrameOrigin_(self.screen.origin)
 
         if window.initial_x is not None and window.initial_y is not None:
-            self.move(window.initial_x, window.initial_y)
+            # initial_x / initial_y are relative to the target screen, move expects absolute
+            screen_top = _primary_screen_height() - self.screen.origin.y - self.screen.size.height
+            self.move(self.screen.origin.x + window.initial_x, screen_top + window.initial_y)
         else:
             self.center()
 
@@ -809,10 +816,7 @@ class BrowserView:
         AppHelper.callAfter(self.window.deminiaturize_, self)
 
     def move(self, x, y):
-        flipped_y = self.screen.size.height - y
-        self.window.setFrameTopLeftPoint_(
-            AppKit.NSPoint(self.screen.origin.x + x, self.screen.origin.y + flipped_y)
-        )
+        self.window.setFrameTopLeftPoint_(AppKit.NSPoint(x, _primary_screen_height() - y))
 
     def center(self):
         window_frame = self.window.frame()
@@ -1569,15 +1573,9 @@ def evaluate_js(script, uid, parse_json=True):
 
 def get_position(uid):
     def _position(coordinates):
-        screen_frame = i.screen
-
-        if screen_frame is None:
-            raise RuntimeError('Failed to obtain screen')
-
-        window = i.window
-        frame = window.frame()
+        frame = i.window.frame()
         coordinates[0] = int(frame.origin.x)
-        coordinates[1] = int(screen_frame.size.height - frame.origin.y - frame.size.height)
+        coordinates[1] = int(_primary_screen_height() - frame.origin.y - frame.size.height)
         semaphore.release()
 
     coordinates = [None, None]


VictorPurMar:
The macOS check shows cancelled, but the tests inside it passed: `126 passed, 3 skipped in 195.53s`. The job then sat idle until the 20 minute `timeout-minutes` cap, leaving two orphan Python processes.

Same signature on two earlier runs on branches unrelated to this change, so it looks pre-existing rather than caused by the patch:

- [25792689765](https://github.com/r0x0r/pywebview/actions/runs/25792689765) (May 13): `126 passed, 3 skipped in 170.29s`, cancelled after 20m29s, two orphan Pythons
- [25522656759](https://github.com/r0x0r/pywebview/actions/runs/25522656759) (May 7, winui3): cancelled after 20m21s

Locally the same suite finishes and exits clean with the patch: `126 passed, 3 skipped in 119.88s` on macOS 26.5.

PR #1803, opened 2026-03-30
Add .NET 8 (coreclr) compatibility for Windows ARM64

smparkes:
pywebview fails to load on Windows ARM64 because pythonnet's .NET Framework (netfx) runtime is not available on ARM64 — the clr-loader dependency chain (NXPorts → dnlib) does not support ARM64: https://github.com/pythonnet/pythonnet/discussions/2523

The upstream fallback in #1791 successfully switches to coreclr when netfx is unavailable, but pywebview then fails because several components assume .NET Framework.

Use FolderBrowserDialog for the folder picker on coreclr, since the internal FileDialogNative types (IFileDialog, FOS_PICKFOLDERS) used by the existing OpenFolderDialog don't exist in .NET 8. On .NET Framework, the existing FileDialogNative reflection code is preserved unchanged, including its support for multiple folder selection. FolderBrowserDialog does not support multi-select; this limitation only affects coreclr users. (.NET 9 adds FolderBrowserDialog.Multiselect which could be used in a future update.) The netfx code is wrapped in an if/else on PYTHONNET_RUNTIME, which adds indentation; ignoring whitespace (https://github.com/r0x0r/pywebview/compare/master...smparkes:fix/dotnet8-coreclr?expand=1&w=1) makes the diff much simpler to review.

Update bundled WebView2 interop DLLs from Microsoft.Web.WebView2 NuGet package 1.0.2957.106 to 1.0.3240.44 (netcoreapp3.0 target): https://www.nuget.org/packages/Microsoft.Web.WebView2/1.0.3240.44 The old DLLs referenced System.Windows.Forms.ContextMenu, which was removed in .NET 6+:
https://learn.microsoft.com/en-us/dotnet/core/compatibility/windows-forms/6.0/apis-throw-argumentnullexception The new DLLs use ContextMenuStrip instead, which exists on both .NET Framework and .NET 8, so this change is safe for all platforms.

Skip the .NET Framework registry check in _is_chromium() when running on coreclr, since coreclr does not register in the .NET Framework registry (HKLM\...\NDP\v4\Full). Also fix a potential UnboundLocalError by initializing net_key before the try block and guarding CloseKey.

Add explicit clr.AddReference('Microsoft.Win32.SystemEvents') when PYTHONNET_RUNTIME=coreclr. This assembly is auto-loaded on .NET Framework but requires an explicit reference on coreclr.

PR #1832, opened 2026-08-11
coreclr with a .NET Framework fallback on Windows

smparkes:
Reviving #1803 with the design @r0x0r asked for there:

> My take is that it makes sense to target coreclr by default and fall back to netfx. Not sure how dramatic change this is, but maybe this can be released in the next major version.

That's what this does. I'd originally proposed selecting by architecture (coreclr on ARM64, netfx elsewhere); this implements @r0x0r's version instead — coreclr wherever it's available, netfx when it isn't.

## Why

- .NET Framework doesn't exist on Windows on ARM64, so the WinForms backend can't start there at all.
- pythonnet defaults to netfx, and each Windows backend only nudged it toward coreclr *after* an import had already failed (`try: import clr / except: PYTHONNET_RUNTIME = 'coreclr'`). On a machine that has .NET Framework, that always lands on netfx.
- The bundled WebView2 assemblies target .NETFramework, so anyone who forces coreclr gets a `TypeLoadException` from them — the first of the three failures @JeffSteinbok reported on #1803.

## What changed

`webview/clr_runtime.py` chooses the runtime: coreclr first, netfx if coreclr is unavailable. An embedder's explicit choice is never overridden — if `PYTHONNET_RUNTIME` is already set, that runtime is used as-is, and an existing `PYTHONNET_CORECLR_RUNTIME_CONFIG`, from an application bundling its own private .NET, is left alone.

The bootstrap itself lives in `webview/platforms/_pythonnet.py`, which exists because winforms, edgechromium and mshtml all need `clr` and all depend on the same easily broken rule: the runtime must be chosen before pythonnet is imported, since pythonnet reads `PYTHONNET_RUNTIME` as it loads and cannot be told afterwards. Each backend previously carried its own copy of that bootstrap. Now one module owns it and the backends import `clr` from there, which also means importing a backend by any path gets the guarantee rather than depending on a caller to have called something first.

## Why a runtimeconfig is involved at all

This is the part I'd never have guessed from the outside, so it's worth spelling out.

On .NET Framework there is no such thing as a runtimeconfig. WinForms is part of the framework, so `clr.AddReference('System.Windows.Forms')` just works and nothing needs configuring. `runtimeconfig.json` is a .NET (Core) mechanism: hostfxr reads it to decide which *shared framework* to load, and WinForms is not in the base `Microsoft.NETCore.App` — it lives in `Microsoft.WindowsDesktop.App`, a separate pack the runtimeconfig has to name.

Which is why the existing coreclr branch could never have worked:

```python
try:
    import clr
except Exception:
    os.environ['PYTHONNET_RUNTIME'] = 'coreclr'
    import clr
```

It names the runtime and stops. With no config to work from, clr_loader synthesizes one for the *base* runtime, so the CLR starts without WinForms — and the next statement in the module is `clr.AddReference('System.Windows.Forms')`. That branch is only reachable on a machine with no .NET Framework, and on precisely those machines it dies a few lines later. So the missing runtimeconfig was never anyone's visible problem: nothing ever got far enough for it to be.

The file this PR adds is a stock .NET runtimeconfig — the same schema `dotnet build` emits next to every application — asking for the desktop pack and rolling forward:

```json
{
  "runtimeOptions": {
    "tfm": "net6.0-windows",
    "rollForward": "LatestMajor",
    "framework": {
      "name": "Microsoft.WindowsDesktop.App",
      "version": "6.0.0"
    }
  }
}
```

`version` is a floor rather than a pin, and `rollForward: LatestMajor` is what lets a request for 6.0.0 be satisfied by 8, 9 or 10. That direction matters: rollForward only ever rolls *up*, so pinning a recent major would fail outright on hosts with an older runtime — which is how @JeffSteinbok's .NET 8/9 machines would have been excluded.

So this isn't configuration the PR invents. It's the piece the coreclr branch was always missing, expressed in .NET's own format and reached through pythonnet's own documented parameter.

`interop_dll_path` then resolves assemblies for whichever runtime is active. It searches several directories in turn: the package's own `webview/lib`, and for frozen applications the directory beside the executable and `sys._MEIPASS`. Under coreclr it now looks in a `netcoreapp3.0/` subdirectory of each one first, falling back to the assemblies sitting directly in that directory, which are the net462 copies. The frozen directories need this as much as the package one — PyInstaller stages `webview/lib` wholesale, so without the same check a bundled coreclr application would load the netfx assemblies staged beside its executable.

## Environment

No new environment variables. `PYTHONNET_RUNTIME` was already set by the old `try: import clr` block. `PYTHONNET_CORECLR_RUNTIME_CONFIG` is pythonnet's own parameter, not something invented here — pythonnet turns any `PYTHONNET_<RUNTIME>_<PARAM>` variable into a keyword argument for the clr_loader factory, so this one arrives as `runtime_config=` on `get_coreclr()`, and its docstring names this exact variable as the one to set for `PYTHONNET_RUNTIME=coreclr`.

The only new *write* is a `setdefault`: pywebview fills it in when nobody else has, so an application bundling its own private .NET keeps its own runtimeconfig. `DOTNET_ROOT` is neither read nor written here.

Both variables have to be set before pywebview initialises a GUI — backend selection is lazy, happening on the first `webview.start()` or `webview.screens` rather than at import — so setting them before `import webview` is the safe rule.

The installation guide and the engine table both claimed .NET Framework was required; those are updated, along with a note on overriding either variable.

## Folder dialogs

This is the one place where switching runtime could have cost a feature. `OpenFolderDialog` reflects on internal `FileDialogNative` types that don't exist on coreclr, so the coreclr build uses `FolderBrowserDialog` instead — but that class only gained `Multiselect` and `SelectedPaths` in .NET 9, while the .NET Framework variant has honoured `allow_multiple` since "Support multiple selection in open folder dialog" landed. Defaulting to coreclr would have quietly taken that away from anyone with both runtimes.

So the substitute probes for the members and uses them when they're there, rather than testing a version, which means it lights up on its own as users move to .NET 9+. Below that it returns a single folder, as `FolderBrowserDialog` has nothing else to offer.

The API docs for `create_file_dialog` said only that `allow_multiple=True` enables multiple selection; they now name the platforms where multiple *folders* work. That is not uniform and wasn't before this PR either — QT resolves folder dialogs through `QFileDialog.getExistingDirectory`, which is single by construction.

## About the added files

Three files land in `webview/lib`:

- `netcoreapp3.0/Microsoft.Web.WebView2.Core.dll` and `.WinForms.dll` — added *alongside* the existing net462 assemblies rather than replacing them. `webview/lib/*.dll` is untouched, so .NET Framework hosts are unaffected and the netfx fallback stays possible. They're the same WebView2 SDK version @r0x0r already updated master to.
- `pywebview-runtimeconfig.json` — the default runtimeconfig handed to clr_loader. It asks for `Microsoft.WindowsDesktop.App`, because WinForms is not part of the base runtime, and rolls forward to whatever major is installed rather than pinning one. An embedder overrides it with `PYTHONNET_CORECLR_RUNTIME_CONFIG` instead of editing it.

## Commits

**Worth reading this diff with whitespace hidden** — tick *Hide whitespace changes*, or append `?w=1` to the diff URL. Gating `OpenFolderDialog` on the runtime re-indents the entire class, so `winforms.py` looks rewritten when it mostly isn't: +121/−71 by default, +56/−6 with whitespace off. The first commit goes from +121/−66 to +56/−1, and the PR as a whole from +458/−100 to +392/−34.

@JeffSteinbok suggested splitting #1803 into per-fix PRs. I've kept them as separate commits rather than separate PRs because individually they don't accomplish anything: the `SystemEvents` fix only matters once coreclr can start, and coreclr can't start until the assemblies and the runtimeconfig are in place. They are curated to be self-contained, so the PR can be reviewed commit by commit. Happy to split them into separate PRs if @r0x0r would rather.

1. `Add .NET 8 (coreclr) compatibility for Windows ARM64` — the #1803 fixes: `Microsoft.Win32.SystemEvents` AddReference, a `FolderBrowserDialog` substitute for `OpenFolderDialog` (the internal `FileDialogNative` types it reflects on don't exist on coreclr), and a `_is_chromium` guard so it doesn't consult the .NET Framework registry key under coreclr. The WebView2 DLL update from the original PR is dropped — that landed on master already.
2. `Ship netcoreapp3.0 WebView2 assemblies alongside the net462 ones`
3. `Prefer coreclr, fall back to .NET Framework` — the selection, the runtimeconfig, per-runtime assembly resolution, the tests, the packaging fix below, and the documentation updates.
4. `Report why coreclr was rejected when falling back to netfx` — the fallback swallowed the coreclr exception, so a host where both fail reported only the netfx error, which on ARM64 is the useless half.

## Testing

`tests/test_clr_runtime.py` covers selection order, the netfx fallback, the error naming both failures, the runtimeconfig rolling forward across majors, and per-runtime resolution including the frozen path. These are pure unit tests with an injected loader — no Windows or .NET required, so CI runs them everywhere.

One thing they can't cover: the runtimeconfig has to be *packaged*, and `webview.lib` package-data globbed only `**/*.dll` and `**/*.jar`, so the first `.json` in that directory was silently absent from the wheel. `**/*.json` is added. Worth knowing that CI installs with `pip install -e .`, which imports from the source tree and never consults package-data, so a wheel-contents check is the only thing that would catch this — I've left adding one out of this PR.

Manually verified for Windows ARM64 and x64 builds (x64 run under emulation under ARM64 Windows), both as a plain package and inside a PyInstaller bundle, driving an application that embeds pywebview: the WebView2 window comes up on both architectures, and folder dialogs open.

Two things that testing does not reach, since I'd rather say so than have them found. The netfx fallback has only ever run in the unit tests — every machine I have has .NET, so the real runs all took the coreclr path. And the folder dialog was smoked with `allow_multiple` unset, which is the single-folder path; the multiselect branch has not run against a real .NET 9 dialog.

@JeffSteinbok — if the offer to test still stands: `pip install git+https://github.com/macguffin-dev/pywebview@pr/coreclr`

Worth noting these patches have been running in a fork for a while — the app they support ships on both Windows architectures — but the runtime *selection* here is new, since the fork previously forced coreclr from the application side.

Happy to hold this for a major release as @r0x0r suggested. The netfx fallback means existing users keep working, but coreclr-by-default is a genuine behaviour change on machines that have both runtimes.


smparkes:
I'll rebase on the post-WinUI3 stuff presently.

smparkes:
* Update to main
* Not too much in the way of changes; doesn't touch the winui3 stuff
* Added a test (tests/test_clr_runtime_live.py) where clr was expected but not correctly configured
* Tested new code in app on macos, linux, windows11, all arm64 but also running x64 under rosetta/prism
* Fixed an existing whitespace issue in the license to clear the PR check

PR #1844, opened 2026-09-11
Fix GUI-thread reentrancy deadlocks across the backends

r0x0r:
Fixes a family of GUI-thread reentrancy deadlocks that existed on every backend, and adds `tests/test_deadlocks.py` to keep them fixed.

## The bug

Four events are dispatched **synchronously on the GUI thread** so a handler can veto or mutate the outcome — the `should_lock=True` events in `webview/window.py`:

```
closing, before_show, before_load, initialized
```

While such a handler runs, it *is* the GUI thread. If it then called a synchronous webview API, the platform layer posted the real work back to the GUI thread and blocked on a semaphore:

```python
# cocoa.BrowserView.evaluate_js, before this PR
AppHelper.callAfter(eval)             # queued onto the (now-blocked) GUI runloop
JSResult.result_semaphore.acquire()   # GUI thread parks here forever
```

The queued work can never run, because the only thread that could run it is parked. The same shape existed on GTK (`glib.idle_add` + `Semaphore.acquire`), WinForms/EdgeChromium (`Invoke` + `ContinueWith(syncContextTaskScheduler)` + `Semaphore.acquire`), Qt and WinUI3.

It is unrecoverable: a blocking `threading` semaphore acquire on the main thread does not poll for signals, so neither `KeyboardInterrupt` nor `_thread.interrupt_main()` breaks it. The process has to be killed.

## The contract

No pywebview API may block when called from the GUI thread. What it does instead depends on what the native operation actually needs:

- **Synchronous native operation → runs inline** on the calling thread and returns its value. On every backend that covers `window.width`/`height`/`x`/`y`, `get_current_url()`, `run_js()` and `window.state` updates, plus the confirmation and file dialogs on Cocoa and GTK, whose panels pump their own modal loop.
- **Asynchronous native operation → raises `ReentrantCallError`.** Its result would be delivered by the very thread that is asking, so it cannot be produced here at all. This covers `evaluate_js()` on every backend and `get_cookies()` on all but Qt, which answers from its own cookie cache.

`webview.errors.ReentrantCallError` subclasses `RuntimeError`, so code already catching the plain `RuntimeError` that GTK and WinUI3 raised at these sites keeps working.

## Per backend

| | before | after |
|---|---|---|
| **Cocoa** | `get_current_url`, `evaluate_js`, `get_cookies`, both dialogs deadlocked; `get_position`/`get_size` guessed with try/except | `_call_on_ui_thread()` helper; reads and dialogs inline, `evaluate_js`/`get_cookies` raise |
| **GTK** | `evaluate_js`/`get_cookies` raised; `get_current_url`, `get_position`, `get_size` and both dialogs raised too | same helper; those five now run inline instead of refusing work they could do |
| **Qt** | `evaluate_js` deadlocked | handled; the other emits already ran their slot inline via Qt's direct connection, so nothing else needed changing |
| **WinForms / EdgeChromium** | `evaluate_js`, `get_cookies` deadlocked | `IsHandleCreated` + `InvokeRequired` guard (`InvokeRequired` alone is False on a handle-less control from *any* thread) |
| **WinUI3** | already correct | moved to the shared error type |

CEF was checked and is not affected: it runs with `multi_threaded_message_loop`, so its JS results arrive on a CEF thread. mshtml is likewise fine — `InvokeScript` is synchronous.

## Tests

`tests/test_deadlocks.py` calls each API **directly from a `closing` handler**, which is the only faithful reproduction. An earlier revision of this branch used a probe thread; that measured the wrong thing, since a background thread waiting on a GUI thread that is merely *busy* running a handler is not a deadlock, it just waits.

Making the real call means a regression hangs rather than fails, so each run is wrapped in a `faulthandler` watchdog. Verified by disabling the fix and re-running: the process dies in seconds with `Timeout (0:00:12)!` and a stack naming the blocked `acquire()` inside `_call_on_ui_thread`, instead of hanging until the job times out. The closer also always closes the window now — skipping it on a load timeout left `start()` waiting forever, which is how one earlier local run wedged for 10 minutes.

## Verification

All 8 CI jobs green. The 10 cases run on Ubuntu GTK, Ubuntu Qt, Windows EdgeChromium and macOS; full suites there are 147, 144, 152 and 148 passed respectively. `pre-commit run` passes. Docs and CHANGELOG updated.

## Two WinUI3 issues found along the way (both pre-existing, neither fixed here)

**1. `closing` never fires for a programmatic `destroy()`.** `on_closing` is attached to `AppWindow.Closing`, which `Microsoft.UI.Xaml.Window.Close()` does not raise. CI confirms it: the window loads, `destroy()` returns cleanly, `start()` finishes — and the handler never ran (`{'handler_ran': False, 'loaded': True, 'closed': True}`). So `closing` handlers and `confirm_close` are both skipped whenever a window is closed from code rather than by the user. Cocoa has the same gap, which is why the test calls `performClose_` there; WinUI3 has no equivalent public API, so the closing-handler cases are skipped with an explicit reason.

**2. A failing test could not turn the job red.** `run.ps1` does sum each pytest exit code, but `Application.current.exit` in the last-window-close teardown tears the process down with status 0 before pytest prints its summary or sets an exit code. Before the skip was added, `test_deadlocks.py` printed a bare `F` into a **green** job. This also explains why only the first test of each window-opening file ever runs on WinUI3 (`test_dom.py`, `test_evaluate_js.py` behave the same today).

Both are worth fixing separately; I did not touch them here since neither is a deadlock.

https://claude.ai/code/session_01SNsfteUN5FpA7GFjE3RQyq


r0x0r:
@copilot please re-review the latest changes addressing your feedback.

